PR #8, opened 2017-03-29
removed absolute path in one test

alfoa:
cdt


alfoa:
Just removed an absolute path. No need to review

Issue #42, opened 2017-04-06
Consistent Module Imports


PaulTalbot-INL:
--------
Issue Description
--------
##### What did you expect to see happen?
Consistent, modular imports throughout RAVEN

##### What did you see instead?
Fiddling with PYTHONPATH and some inconsistency in when some modules can import others safely, as well as occasional imports in the body of python modules

##### Do you have a suggested fix for the development team?
We should probably continue towards making RAVEN more library-like, so that our imports all happen in a consistent way.

----------------
For Devs: Issue Review
----------------
- [x] 1. Is it tagged with a type: defect or improvement? __tagged__
- [x] 2. Is it tagged with a priority: critical, normal or minor? __tagged__
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements? __n/a__
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users. __n/a__
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.) __provided__

-------
For Devs: Issue Closure
-------
- [x] 1. If the issue is a defect, is the defect fixed?
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [x] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)?
- [x] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided?

joshua-cogliati-inl:
Note that when this is fixed, we can probably add some useful checks to the pylint checks in developer_tools/pylint_list  and developer_tools/pylint_check

joshua-cogliati-inl:
I think this might have gotten fixed at somepoint, but never closed.  As in I think the great renaming of ```import utils``` to ```import utils.utils``` fixed this.

joshua-cogliati-inl:
Hm, fixing this might be required for #1764 to get a reasonable wheel created.

PR #131, opened 2017-04-25
Parallel Input Generation and Output Collection for MultiRun Steps

### LINKED ISSUE #130: Parallel Input Generation and Output Collection for MultiRun steps
--------
Issue Description
--------

I expect the inputs of a multi-run to be generated in parallel, and in addition, I expect the outputs of each run to be read and collected in parallel.

Currently, RAVEN will do both of these operations in serial on the main thread.

##### Do you have a suggested fix for the development team?

I am working on a development that will move the `createNewInput` and `finalizeModelOutput` functions to occur on the parallel jobs. This is a bit of an invasive change as it changes how data is fundamentally passed around on the inside of RAVEN and so should be tested and reviewed thoroughly before acceptance.

##### Please attach the input file(s) that generate this error.  The simpler the input, the faster we can find the issue.

This is a new improvement that will affect all existing test cases that utilize the `MutiRun` step type.

----------------
----------------
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements?  __Should not impact requirements, only increases efficiency of data handling__
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.  __Not a defect__

-------
-------
- [ ] 1. If the issue is a defect, is the defect fixed? ***improvement***
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.) ***improvement***
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)? ***improvement***
- [x] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)? ***improvement***
- [x] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided? ***#131***

maljovec:
--------
Pull Request Description
--------

Closes #130.

This PR moves the creation of inputs and finalization of output files for all model types to parallel jobs handled by the JobHandler. This also changes how each job submitter must create a "realization" and get information about the returned data. This makes significant changes throughout the code and should be reviewed by multiple developers to ensure compliance and also to ensure that developers are aware of the new data pathways brought out by this PR.

----------------
----------------


framework/CodeInterfaces/Generic/GenericParser.py
```diff
@@ -52,7 +52,8 @@ def __init__(self,inputFiles,prefix='$RAVEN-',postfix='$',defaultDelim=':', form
       infileName = inputFile.getFilename()#os.path.basename(inputFile)
       self.segments[infileName] = []
       if not os.path.exists(inputFile.getAbsFile()):
-        raise IOError('Input file not found: '+inputFile)
+        ## Make sure to cast the inputFile to a string as it may be File object.
+        raise IOError('Input file not found: ' + str(inputFile))
```
PaulTalbot-INL: FYI, if a comma is used instead of concatenating the arguments, it does cast it as a string.  No change required.

framework/Models.py
```diff
@@ -1468,59 +1451,81 @@ def createNewInput(self,currentInput,samplerType,**Kwargs):
...
-    """
-      Method that performs the actual run of the Code model
-      @ In,  Input, object, object contained the data to process. (inputToInternal output)
-      @ In,  jobHandler, JobHandler instance, the global job handler instance
-      @ Out, None
+
+    kwargs['subDirectory'] = subDirectory
+
+    if 'SampledVars' in kwargs.keys():
+      sampledVars = self._replaceVariablesNamesWithAliasSystem(kwargs['SampledVars'],'input',False)
```
PaulTalbot-INL: Is this something we do for all models?  It might be abstracted in the future.  No change necessary now.

framework/Models.py
```diff
@@ -1468,59 +1451,81 @@ def createNewInput(self,currentInput,samplerType,**Kwargs):
...
+    ## This could cause issues as the temporary changing of the directory on
+    ## this thread could affect the main thread's working directory, I would
+    ## prefer to copy the environment and then update the working directory of
+    ## the new environment, but I am not sure if this is OS agnostic? --DPM 4/25/17
+    ## push directory
+    # oldDir = os.getcwd()
+    # os.chdir(metaData['subDirectory'])
+
+    ## This is the preferred alternative, but we should ensure that it works
+    ## on all systems
```
PaulTalbot-INL: With the tests passing, do we have reasonable confidence this works on all systems now to the extent we can update these comments?

framework/Optimizers/Optimizer.py
```diff
@@ -580,13 +585,15 @@ def handleFailedRuns(self,failedRuns):
...
         self.raiseADebug('      return code :',run.getReturnCode())
-        if metadata is not None:
-          self.raiseADebug('      sampled vars:')
-          for v,k in metadata['SampledVars'].items():
-            self.raiseADebug('         ',v,':',k)
+        # metadata = run.getMetadata()
+        # if metadata is not None:
+        #   self.raiseADebug('      sampled vars:')
+        #   for v,k in metadata['SampledVars'].items():
+        #     self.raiseADebug('         ',v,':',k)
```
PaulTalbot-INL: should this block of commented code be removed?

framework/Runners/__init__.py
```diff
@@ -25,7 +25,7 @@
 ## 'from Runners.Runner import Runner' outside
 ## of this submodule
 from .Runner import Runner
-from .ExternalRunner import ExternalRunner
+# from .ExternalRunner import ExternalRunner
```
PaulTalbot-INL: This doesn't exist anymore, right?  Is it a good idea to remove the commented code?

framework/CodeInterfaces/Generic/GenericParser.py
maljovec: So, you are saying the line could be:

```
raise IOError('Input file not found: ', inputFile)
```
and it would work as well? This is good to know as this is slightly shorter syntax and cleaner.

PaulTalbot-INL: Oh, this is my mistake, the message handler isn't involved here.  I'm completely wrong on this, you were right originally.

maljovec: Okay, but still good to know for the messageHandler. I probably still would have used the long-form syntax. 👍

framework/Optimizers/Optimizer.py
maljovec: This could probably be left in the code on second thought. Given other development efforts, it is probably best for me not to unnecessarily touch parts of the code like this. I will unroll this change when I receive comments from everyone else as long as it does not break anything (it should not).

framework/Runners/__init__.py
maljovec: Yes, this vestigial piece should be removed from this file so as not to promote confusion later on. Good catch!

framework/JobHandler.py
```diff
@@ -626,19 +552,34 @@ def addRuns(self):
...
+            ## to replace the execution command. Is this fragile? Possibly. We may
+            ## want to revisit this on the next iteration of this code.
+            if len(item.args) > 0 and isinstance(item.args[0], Models.Code):
+              kwargs = {}
+              kwargs['INDEX'] = str(i)
+              kwargs['INDEX1'] = str(i+i)
+              kwargs['CURRENT_ID'] = str(self.__nextId)
+              kwargs['CURRENT_ID1'] = str(self.__nextId+1)
+              kwargs['SCRIPT_DIR'] = self.runInfoDict['ScriptDir']
+              kwargs['FRAMEWORK_DIR'] = self.runInfoDict['FrameworkDir']
```
alfoa: I know that this is not pertinent to this PR, but this wildcard replecament should be done at the Simulation level.

The reason is that, if we do that at this stage, we con not check for the "existence" of the executable that the user inputs (And this is the reason why when we testify the not existence of the executable we just raise a message and not an error (as we should do )):
```
      if os.path.exists(abspath):
        self.executable = abspath
      else:
        self.raiseAMessage('not found executable '+self.executable,'ExceptedError')
```

joshua-cogliati-inl: I am not sure that we can check that the executable exists at XML read time and make it an error.  For example, if we are reading the xml on the head node, the executable might not exist there and only be found on the run nodes.

framework/Models.py
```diff
@@ -1468,59 +1451,81 @@ def createNewInput(self,currentInput,samplerType,**Kwargs):
...
+    sampleDirectory = os.path.join(os.getcwd(),metaData['subDirectory'])
+    localenv = dict(os.environ)
+    localenv['PWD'] = str(sampleDirectory)
```
joshua-cogliati-inl: Hm, we are setting PWD = sampleDirectory, but cwd=metaData['subDirectory'].  Is this correct and does this make sense?

framework/Models.py
```diff
@@ -1753,19 +1989,20 @@ def collectOutput(self,finishedjob,output,options=None):
...
       @ In, myInput, list, the inputs (list) to start from to generate the new one
       @ In, samplerType, string, is the type of sampler that is calling to generate a new input
-      @ In, **Kwargs, dict,  is a dictionary that contains the information coming from the sampler,
+      @ In, **kwargs, dict,  is a dictionary that contains the information coming from the sampler,
            a mandatory key is the sampledVars'that contains a dictionary {'name variable':value}
       @ Out, createNewInput, tuple, return the new input in a tuple form
     """
-    return self.interface.inputToInternal(myInput)
+    # return self.interface.inputToInternal(myInput)
+    return myInput
```
alfoa: docstrings is not valid anymore...

framework/Models.py
```diff
@@ -2151,17 +2390,50 @@ def getAdditionalInputEdits(self,inputInfo):
...
+        @ In, samplerType, string, is the type of sampler that is calling to generate a new input
+        @ In, **kwargs, dict,  is a dictionary that contains the information coming from the sampler,
+           a mandatory key is the sampledVars'that contains a dictionary {'name variable':value}
+        @ Out, returnValue, dict, This holds the output information of the evaluated sample.
+    """
+    jobHandler = kwargs.pop('jobHandler')
+    Input = self.createNewInput(myInput[0], samplerType, **kwargs)
+
+    ## Unpack the specifics for this class, namely just the jobHandler
+    return (Input,self._externalRun(Input,jobHandler))
```
alfoa: docstrings (please check all of them)

framework/Runners/ExternalRunner.py
```diff
@@ -1,274 +0,0 @@
...
-    """
-    return self.workingDir
-
-  def getOutputFilename(self):
-    """
-      Function to get the output filenames
-      @ In, None
-      @ Out, self.output, string, output filename root
-    """
-    return os.path.join(self.workingDir,self.output)
```
alfoa: I understand that this is done internally now, but why we decided to remove it? Wouldn't it be more "object-oriented" keeping this approach? No change requested here, but just an explanation why the current implementation (internalization in the Models) is better that this approach.

framework/Samplers/DynamicEventTree.py
```diff
@@ -183,7 +183,27 @@ def localFinalizeActualSampling(self,jobObject,model,myInput,genRunQueue=True):
...
+    ## There are two ways to get at the working directory from the job instance
+    ## and both feel a bit hacky and fragile to changes in the Runner classes.
+    ## They are both listed below and the second inevitably stems from the first.
+    ## I am wondering if there is a reason we cannot use the model.workingDir
+    ## from above here? Granted the job instance should have a snapshot of
+    ## whatever the model's current working directory was for that evaluation,
+    ## and it could have changed in the meantime, so I will keep this as is for
+    ## now, but note this should be re-evaluated in the future. -- DPM 4/12/17
+    # codeModel = jobObject.args[0]
+    # jobWorkingDir = codeModel.workingDir
```
alfoa: ```jobObject.getWorkingDir()``` can not be used anymore?

framework/Models.py
maljovec: I agree with this observation.  It turns out `sampleDirectory` is just the absolute directory of what `metaData['subDirectory']` holds, so we should be safe to have cwd=sampleDirectory without any ill effects. I will make this change locally and ensure that it does not break anything else.

maljovec: Actually, I will set cwd=localenv['PWD'] to future proof this and ensure that they stay synced. (setting cwd may not be necessary?)

framework/JobHandler.py
alfoa: Can I have an example of this? Generally what is in the head node is available (at least in terms of executable) in the run nodes. What is the exception? 

Probably "alias" definitions?

framework/Models.py
joshua-cogliati-inl: If I had to guess, I am betting that setting PWD is not necessary, and setting cwd is very necessary.

framework/Runners/ExternalRunner.py
maljovec: I thought this might come up and I had this debate internally as well. My reasoning was: does it make sense to have a piece of unused code sitting in the repository that will slowly become out of sync with updates made to the rest of the code in the event that we will want it back in the future?  My answer was no, the code still exists in the revision history and a developer will be to extract it by digging it out, but they will be aware that the code they are dealing with worked on a __previous__ version of RAVEN.

By virtue of leaving it in the repository, we need to maintain it as we do everything else which means writing hypothetical tests to test a hypothetically useful piece of code. The majority of the inner workings of this code have not been thrown away, but instead moved.

Sorry, this was a bit of a digression, I contemplated deleting this post, but I still think it is useful information for others to see.

maljovec: The other option is to change the Code model to call create an ExternalRunner, which I believe is what you are suggesting, but then we have to places in the code managing "Runners" which seemed ripe for confusion to me.

framework/JobHandler.py
joshua-cogliati-inl: Hm, the code above in Model.py has another problem:
```
abspath = os.path.abspath(str(self.executable))
```
will work badly for executables that are on the path.  As in:
```
>>> os.path.abspath("ls")
'/home/joshcogliati/ls'
```
It may not be very common, but the NCSU cluster had different locations for some of the python  libraries on the head node and the cluster nodes.

framework/Runners/ExternalRunner.py
maljovec: We could have the Code create an ExternalRunner, but it started to feel like the ExternalRunner was just an added layer to the Code model. The ExternalRunner and the InternalRunner in my mind interacted very differently in the our code and so I wasn't convinced it was the right hierarchy to take. I admit I could be wrong on this.

Only the Code used the ExternalRunner.

As another point, the JobHandler did a lot of extra checking to see if it was dealing with an ExternalRunner. This didn't feel clean to me.  In an object-oriented approach, the JobHandler wouldn't care what kind of Runner it was dealing with it and it would allow dynamic typing to ensure that each Runner was processed correctly.

framework/Models.py
```diff
@@ -1529,44 +1534,197 @@ def run(self,inputFiles,jobHandler):
...
       @ In, finishedJob, InternalRunner object, instance of the run just finished
       @ In, output, "DataObjects" object, output where the results of the calculation needs to be stored
       @ In, options, dict, optional, dictionary of options that can be passed in when the collect of the output is performed by another model (e.g. EnsembleModel)
       @ Out, None
     """
-    outputFilelocation = finishedjob.getWorkingDir()
-    attributes={"inputFile":self.currentInputFiles,"type":"csv","name":os.path.join(outputFilelocation,finishedjob.output+'.csv')}
-    attributes['alias'] = self.alias
-    metadata = finishedjob.getMetadata()
+    if finishedJob.getEvaluation() == -1:
```
joshua-cogliati-inl: Is -1 the only possible error output?  (If this is from a subprocess, any negative number will be an error code)

maljovec: Actually, this specifically comes from how we return info in the Runner and InternalRunner classes. There an error is given specifically as -1. This value is not tied to the subprocess or thread return value directly. It is processed by the Runner classes.

framework/Models.py
maljovec: Well, the problem is if we don't set PWD in the copied environment it will actually point to the wrong directory, so that must be specified.  I am testing what happens if we omit the cwd parameter from the call.

framework/Models.py
maljovec: The problem is since a Runner can execute any python method, we have no guarantee that -1 is not a valid output here except that everything that is valid is typically a tuple of inputs and outputs, but this is less of a contract and more of a convention currently being used.

Maybe, this interaction needs to be more strictly defined and have less interaction from the Runners. Say, for instance that the Runner always returns whatever the thing it is running returns and the caller should know how to process that data?  In this case, I would agree with you that maybe -1 is not a valid answer, but I think we need to better nail down the contractual agreement of things that can be executed by a Runner in terms of what goes in and what comes out?

framework/Models.py
```diff
@@ -297,6 +299,21 @@ def _handleInput(self, paramInput):
...
+        are needed by createNewInput and thus descriptions are copied from there.
+        @ In, myInput, list, the inputs (list) to start from to generate the new one
+        @ In, samplerType, string, is the type of sampler that is calling to generate a new input
+        @ In, **kwargs, dict,  is a dictionary that contains the information coming from the sampler,
+           a mandatory key is the sampledVars'that contains a dictionary {'name variable':value}
+        @ Out, returnValue, dict, This holds the output information of the evaluated sample.
+    """
+    # inRun = self.createNewInput(myInput, samplerType, **kwargs)
+    # output = self.run(inRun)
+    pass
```
maljovec: Commented code should be removed.

maljovec: Right, the abstract method has no return value because you cannot call an abstract method, but everyone who implements it should have a returnValue that is a dict that holds the output information of the evaluated sample. If we don't specify this here then any new subclasses that get implemented run the risk of violating this contract, or people who use one or more subclasses will not know what to expect without looking at one or more subclasses. 

I believe this is generic information that should be available at the most generic level, but if you have a counter example I would be open to discussion.

framework/Models.py
joshua-cogliati-inl: Hm, in that case maybe we should make an error output class (something like Runner.Error) and return it for an error. We would be a lot less likely to stumble on that than -1.

framework/Models.py
```diff
@@ -492,75 +509,51 @@ def _inputToInternal(self,dataIN,full=False):
...
-      self._replaceVariablesNamesWithAliasSystem(inputOut[1]["SampledVarsPb"],'input',False)
-    return inputOut
-
-  def run(self,Input,jobHandler):
-    """
-      This method executes the model .
-      @ In,  Input, object, object contained the data to process. (inputToInternal output)
-      @ In,  jobHandler, JobHandler instance, the global job handler instance
-      @ Out, None
+    if 'SampledVars' in kwargs.keys() and len(self.alias['input'].keys()) != 0:
```
maljovec: Sounds good. I will add this change to my current local revision with this explanation. Good catch 👍

framework/Models.py
```diff
@@ -492,75 +509,51 @@ def _inputToInternal(self,dataIN,full=False):
...
-        This method is the one is going to be submitted through the jobHandler
-        @ In, inRun, dict, the input
-        @ In, prefix, string, the string identifying this job
-        @ Out, lambdaReturnOut, dict, the return dictionary
-      """
-      return {'OutputPlaceHolder':np.atleast_1d(np.float(prefix))}
-
-    uniqueHandler = Input[1]['uniqueHandler'] if 'uniqueHandler' in Input[1].keys() else 'any'
-    jobHandler.addInternal((inRun,Input[1]['prefix']),lambdaReturnOut,str(Input[1]['prefix']),metadata=Input[1], modulesToImport = self.mods, uniqueHandler=uniqueHandler)
+    return (inRun,{'OutputPlaceHolder':np.atleast_1d(np.float(Input[1]['prefix']))})
```
maljovec: Ah yes! You are right, it is now a tuple. I will be sure to address this for each of the Model classes and the base.

framework/Models.py
```diff
@@ -572,13 +565,17 @@ def collectOutput(self,finishedJob,output,options=None):
...
       self.raiseAnError(AttributeError,"No available Output to collect")
-    evaluation = finishedJob.getEvaluation()
-    if type(evaluation[1]).__name__ == "tuple":
-      outputeval = evaluation[1][0]
+
+    sampledVars,outputDict = finishedJob.getEvaluation()
+
+    if type(outputDict).__name__ == "tuple":
+      outputeval = outputDict[0]
     else:
```
maljovec: Adding to my next push now.

framework/Models.py
```diff
@@ -951,27 +939,30 @@ def evaluate(self,request):
...
+        @ In, samplerType, string, is the type of sampler that is calling to generate a new input
+        @ In, **kwargs, dict,  is a dictionary that contains the information coming from the sampler,
+           a mandatory key is the sampledVars'that contains a dictionary {'name variable':value}
+        @ Out, returnValue, dict, This holds the output information of the evaluated sample.
     """
+    Input = self.createNewInput(myInput, samplerType, **kwargs)
     inRun = self._manipulateInput(Input[0])
-    uniqueHandler = Input[1]['uniqueHandler'] if 'uniqueHandler' in Input[1].keys() else 'any'
-    jobHandler.addInternal((inRun,), self.__externalRun, str(Input[1]['prefix']), metadata=Input[1], modulesToImport=self.mods, uniqueHandler=uniqueHandler)
+    return inRun,self._externalRun(inRun)
```
maljovec: Fixing now, should be adjusted in next push.

framework/Models.py
```diff
@@ -1041,47 +1032,32 @@ def initialize(self,runInfo,inputs,initDict=None):
...
+      extCreateNewInput = self.sim.createNewInput(self,myInput,samplerType,**kwargs)
       if extCreateNewInput== None:
         self.raiseAnError(AttributeError,'in external Model '+self.ModuleToLoad+' the method createNewInput must return something. Got: None')
-      if 'SampledVars' in Kwargs.keys() and len(self.alias['input'].keys()) != 0:
-        Kwargs['SampledVars'] = sampledVars
-      newInput = ([(extCreateNewInput)],copy.deepcopy(Kwargs))
-      #return ([(extCreateNewInput)],copy.deepcopy(Kwargs)),copy.copy(modelVariableValues)
+      if 'SampledVars' in kwargs.keys() and len(self.alias['input'].keys()) != 0:
+        kwargs['SampledVars'] = sampledVars
+      newInput = ([(extCreateNewInput)],copy.deepcopy(kwargs))
```
maljovec: Fixing now, should be addressed in next push.

framework/Runners/ExternalRunner.py
alfoa: thank you for the explanation

framework/Models.py
```diff
@@ -1041,47 +1032,32 @@ def initialize(self,runInfo,inputs,initDict=None):
...
+      #return ([(extCreateNewInput)],copy.deepcopy(kwargs)),copy.copy(modelVariableValues)
```
maljovec: Getting rid of it! ✂️

framework/Models.py
```diff
@@ -1168,16 +1144,19 @@ def __externalRun(self, Input, modelVariables):
...
     """
-       Method that performs the actual run of the imported external model
-       @ In,  Input, object, object contained the data to process. (inputToInternal output)
-       @ In,  jobHandler, JobHandler instance, the global job handler instance
-       @ Out, None
+        This will evaluate an individual sample on this model. Note, parameters
+        are needed by createNewInput and thus descriptions are copied from there.
+        @ In, myInput, list, the inputs (list) to start from to generate the new one
+        @ In, samplerType, string, is the type of sampler that is calling to generate a new input
+        @ In, **kwargs, dict,  is a dictionary that contains the information coming from the sampler,
```
maljovec: Updating for all instances of evaluateSample.

framework/Models.py
```diff
@@ -1168,16 +1144,19 @@ def __externalRun(self, Input, modelVariables):
...
+           a mandatory key is the sampledVars'that contains a dictionary {'name variable':value}
+        @ Out, returnValue, dict, This holds the output information of the evaluated sample.
     """
+    Input = self.createNewInput(myInput, samplerType, **kwargs)
     inRun = copy.copy(self._manipulateInput(Input[0][0]))
-    uniqueHandler = Input[0][1]['uniqueHandler'] if 'uniqueHandler' in Input[0][1].keys() else 'any'
-    jobHandler.addInternal((inRun,Input[1],),self.__externalRun,str(Input[0][1]['prefix']),metadata=Input[0][1], modulesToImport = self.mods,uniqueHandler=uniqueHandler)
+    return (inRun,self._externalRun(inRun,Input[1],))
```
maljovec: Fixed in next push.

framework/Models.py
maljovec: Good question. This did come from the existing code and was only displaced by my edits. I am seeing another piece of code right below this that @alfoa had already commented on some place else, so it would appear there is some redundancy in the separate Models class.  Maybe, I should open an issue regarding some repeated use of different snippets of code in this file?

maljovec: But, looking again, the other code was in a different context, so maybe there is no redundancy in this particular case.

framework/Models.py
maljovec: Good point, this has been tested at this point. I can remove this comment and the surrounding commented code.

framework/Models.py
maljovec: It looks like removing cwd did cause the cluster tests to break, so I will ensure that it has the same info as PWD.

framework/Models.py
maljovec: The evaluateSample docstrings have all been checked and will be updated in next push.

framework/Models.py
maljovec: Fixed.

framework/Samplers/DynamicEventTree.py
maljovec: Only one type of job specified this information and it was the ExternalRunner. InternalRunner jobs do not know have a working directory.

maljovec: Even if the ExternalRunner was kept in the code, my suggestion would be to have it called inside the Code model which means that an InternalRunner would still be wrapped around it, and every jobObject here would be an instance of an InternalRunner and not an ExternalRunner.

maljovec: Because each job now does some work in python before and after each "run," it would be hard to have an ExternalRunner instance that sits next to InternalRunner instances in the jobQueue. The ExternalRunner would have to then know how to generate an input file from a model, and collect output from a model, and thus would only be able to run Models and not arbitrary bits of code.

In the way we have it now, the InternalRunners can run anything they want in parallel, not just models, as evidenced by how they are called in the SparseGrid code.

framework/Runners/ExternalRunner.py
maljovec: Sorry, it got a bit long-winded. I hope it makes sense. If we can do it better, I wouldn't be opposed to another iteration of improvement to this redesign. Some of these things were cobbled together as I went, and on an as-needed basis. I may be lacking vision on some longer term aspects of RAVEN however.

I put a comment below in the DynamicEventTree that further discusses some of this that kind of describes one of the problems I ran into with having an ExternalRunner that sits next to the InternalRunner.  It may be that we can use the ExternalRunner in different contexts?  I don't know, if we think eventually something other than Codes will need to generate subprocesses, then it may be useful to keep revisit whether an ExternalRunner can be used.

framework/Models.py
maljovec: Ooh, I like that idea. I can make a more or less empty class for now that is used here. The question is does this make sense to inherit from Exception or just object?  In this context, we are not going to raise an exception, but rather just return it and test if it is of this type in the caller.

maljovec: FWIW, @PaulTalbot-INL votes for Exception, and I tend to agree at this point. It shouldn't matter much as the file is more or less empty at this point and can be molded as necessary later.

framework/JobHandler.py
maljovec: For my own benefit, does this warrant any changes for this PR?

framework/Models.py
alfoa: As far as I remember there should not be any redundancy on this. (But I can be possibly mistaken)

framework/JobHandler.py
alfoa: no. It is just a side discussion

framework/Models.py
joshua-cogliati-inl: I agree with Paul that Exception seems reasonable (and this is not hard to change, so it is not a long term design decision).

alfoa: ok to me

alfoa:
----------------
----------------
- [ ] 1. Review all computer code. ***3 reviewers (included me) did that. Ok***
- [ ] 2. If any changes occur to the input syntax, there must be an accompanying change to the user manual and xsd schema. If the input syntax change deprecates existing input files, a conversion script needs to be added (see Conversion Scripts). ***No. Internal performance improvement***
- [ ] 3. Make sure the Python code and commenting standards are respected (camelBack, etc.) - See on the [wiki](https://github.com/idaholab/raven/wiki/RAVEN-Code-Standards#python) for details. ***Ok***
- [ ] 4. Automated Tests should pass, including run_tests, pylint, manual building and xsd tests. If there are changes to Simulation.py or JobHandler.py the qsub tests must pass. ***Ok***
- [ ] 5. If significant functionality is added, there must  be tests added to check this. Tests should cover all possible options.  Multiple short tests are preferred over one large test. If new development on the internal JobHandler parallel system is performed, a cluster test must be added setting, in <RunInfo> XML block, the node ```<internalParallel>``` to True. ***N/A***
- [ ] 6. If the change modifies or adds a requirement or a requirement based test case, the Change Control Board's Chair or designee also needs to approve the change.  The requirements and the requirements test shall be in sync. ***N/A***
- [ ] 7. The merge request must reference an issue.  If the issue is closed, the issue close checklist shall be done. ***#130***
- [ ] 8. If an analytic test is changed/added is the the analytic documentation updated/added? ***N/A***

alfoa:
Check list passed...

Merging

Issue #71, opened 2017-04-06
Update support for Python 3


joshua-cogliati-inl:
--------
Issue Description
--------
Much of RAVEN works with Python 3, but parts such cloud pickle and pp do not work with Python3.  

----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [x] 1. Is it tagged with a type: defect or improvement?
- [x] 2. Is it tagged with a priority: critical, normal or minor?
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements? __No impact__
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.  __Not a defect__
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [x] 1. If the issue is a defect, is the defect fixed?
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [x] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)?
- [x] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided?

PaulTalbot-INL:
Approved to close via #896. An email to the user base is not necessary, as all pertinent notifications should come as part of the next release along with an announcement of migration to python 3.

PR #192, opened 2017-05-19
Fix python3

joshua-cogliati-inl:
--------
Pull Request Description
--------
Refs #71 

This adds some support for Python3.

In order to use, the following needs to be done to compile it:
```
export CROW_USE_PYTHON3=TRUE; export CHECK_PYTHON3=1
make 
python3 ./setup3.py build_ext build install --install-platlib=./framework/contrib/AMSC
```

This is not full python3 support, but with this, more tests pass with python3.

----------------
----------------
- [x] 2. If any changes occur to the input syntax, there must be an accompanying change to the user manual and xsd schema. If the input syntax change deprecates existing input files, a conversion script needs to be added (see Conversion Scripts). __No syntax changes__
- [x] 5. If significant functionality is added, there must  be tests added to check this. Tests should cover all possible options.  Multiple short tests are preferred over one large test. If new development on the internal JobHandler parallel system is performed, a cluster test must be added setting, in <RunInfo> XML block, the node ```<internalParallel>``` to True. __Changes support Python 3, so no new functionality in this PR__
- [x] 6. If the change modifies or adds a requirement or a requirement based test case, the Change Control Board's Chair or designee also needs to approve the change.  The requirements and the requirements test shall be in sync.  __No change to requirements tests__
- [x] 7. The merge request must reference an issue.  If the issue is closed, the issue close checklist shall be done.  __References #71__
- [x] 8. If an analytic test is changed/added is the the analytic documentation updated/added?  __No change__

### DIFF
--- a/crow/setup3.py
+++ b/crow/setup3.py
@@ -14,11 +14,11 @@
 from distutils.core import setup, Extension
 import subprocess
 try:
-  eigen_flags = subprocess.check_output(["./scripts/find_eigen.py"])
+  eigen_flags = subprocess.check_output(["./scripts/find_eigen.py"]).decode("ascii")
 except:
   eigen_flags = ""
 include_dirs=['include/distributions','include/utilities','contrib/include']
-if eigen_flags.startswith(b"-I"):
+if eigen_flags.startswith("-I"):
   include_dirs.append(eigen_flags[2:].rstrip())
 swig_opts=['-c++','-py3','-Iinclude/distributions','-Iinclude/utilities']
 extra_compile_args=['-std=c++11']
--- a/crow/setup3.py
+++ b/crow/setup3.py
@@ -18,7 +18,7 @@ try:
 except:
   eigen_flags = ""
 include_dirs=['include/distributions','include/utilities','contrib/include']
-if eigen_flags.startswith("-I"):
+if eigen_flags.startswith(b"-I"):
   include_dirs.append(eigen_flags[2:].rstrip())
 swig_opts=['-c++','-py3','-Iinclude/distributions','-Iinclude/utilities']
 extra_compile_args=['-std=c++11']
@@ -26,7 +26,7 @@ ext = 'py3'
 setup(name='crow',
       version='0.8',
       ext_package='crow_modules',
-      ext_modules=[Extension('_distribution1D'+ext,['crow_modules/distribution1D'+ext+'.i','src/distributions/distribution.C','src/utilities/MDreader.C','src/utilities/inverseDistanceWeigthing.C','src/utilities/microSphere.C','src/utilities/NDspline.C','src/utilities/ND_Interpolation_Functions.C','src/distributions/distributionNDBase.C','src/distributions/distributionNDNormal.C','src/distributions/distributionFunctions.C','src/distributions/DistributionContainer.C','src/distributions/distribution_1D.C','src/distributions/randomClass.C','src/distributions/distributionNDCartesianSpline.C'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args),
-                   Extension('_interpolationND'+ext,['crow_modules/interpolationND'+ext+'.i','src/utilities/ND_Interpolation_Functions.C','src/utilities/NDspline.C','src/utilities/microSphere.C','src/utilities/inverseDistanceWeigthing.C','src/utilities/MDreader.C','src/distributions/randomClass.C'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args)],
+      ext_modules=[Extension('_distribution1D'+ext,['crow_modules/distribution1D'+ext+'.i','src/distributions/distribution.cxx','src/utilities/MDreader.cxx','src/utilities/inverseDistanceWeigthing.cxx','src/utilities/microSphere.cxx','src/utilities/NDspline.cxx','src/utilities/ND_Interpolation_Functions.cxx','src/distributions/distributionNDBase.cxx','src/distributions/distributionNDNormal.cxx','src/distributions/distributionFunctions.cxx','src/distributions/DistributionContainer.cxx','src/distributions/distribution_1D.cxx','src/distributions/randomClass.cxx','src/distributions/distributionNDCartesianSpline.cxx'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args),
+                   Extension('_interpolationND'+ext,['crow_modules/interpolationND'+ext+'.i','src/utilities/ND_Interpolation_Functions.cxx','src/utilities/NDspline.cxx','src/utilities/microSphere.cxx','src/utilities/inverseDistanceWeigthing.cxx','src/utilities/MDreader.cxx','src/distributions/randomClass.cxx'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args)],
       py_modules=['crow_modules.distribution1D'+ext,'crow_modules.interpolationND'+ext],
       )
--- a/framework/CustomCommandExecuter.py
+++ b/framework/CustomCommandExecuter.py
@@ -43,4 +43,4 @@ def execCommand(commandString,self=None,object=None):
     @ In, object, instance, optional, object instance
     @ Out, None
   """
-  execCommandReturn(commandString,self,object)
+  exec(commandString)
--- a/framework/DataObjects/Data.py
+++ b/framework/DataObjects/Data.py
@@ -33,6 +33,7 @@ import abc
 import ast
 import copy
 import numpy as np
+import itertools
 from scipy import spatial
 import xml.etree.ElementTree as ET
 #External Modules End--------------------------------------------------------------------------------
@@ -639,7 +640,7 @@ class Data(utils.metaclass_insert(abc.ABCMeta,BaseType)):
     else:
       if typeVar.lower() in ['input','inputs','unstructuredInput']:
         returnDict = {}
-        if keyword in self._dataContainer['inputs'].keys() + self._dataContainer['unstructuredInputs'].keys():
+        if keyword in itertools.chain(self._dataContainer['inputs'].keys(),self._dataContainer['unstructuredInputs'].keys()):
           returnDict[keyword] = {}
           if self.type == 'HistorySet':
             for key in self._dataContainer['inputs'][keyword].keys():
--- a/framework/Samplers/Sampler.py
+++ b/framework/Samplers/Sampler.py
@@ -295,7 +295,7 @@ class Sampler(utils.metaclass_insert(abc.ABCMeta,BaseType),Assembler):
           self.raiseAnError(IOError,'The following are duplicated variables indices listed in the manifestVariablesIndex: ' + str(dups))
         listElement = self.distributions2variablesMapping[dist]
         for var in listElement:
-          self.variables2distributionsMapping[var.keys()[0]]['totDim'] = maxDim #reset the totDim to reflect the totDim of original input space
+          self.variables2distributionsMapping[utils.first(var.keys())]['totDim'] = maxDim #reset the totDim to reflect the totDim of original input space
         tempListElement = {k.strip():v for x in listElement for ks,v in x.items() for k in list(ks.strip().split(','))}
         listIndex = []
         for var in listLatentElement:
--- a/framework/utils/TreeStructure.py
+++ b/framework/utils/TreeStructure.py
@@ -321,30 +321,34 @@ def inputTreeToGetpot(ts,fromNode=False):
 ###########
 # PARSERS #
 ###########
-class XMLCommentParser(ET.XMLTreeBuilder):
-  """
-    An XML parser that expands on the default to preserves comments
-  """
-  def __init__(self):
+if sys.version_info.major > 2:
+  #In Python3, XMLParser is a C implementation, and can't be modified.
+  XMLCommentParser = ET.XMLParser
+else:
+  class XMLCommentParser(ET.XMLParser):
     """
-      Constructor.
-      @ In, None
-      @ Out, None
+      An XML parser that expands on the default to preserves comments
     """
-    ET.XMLTreeBuilder.__init__(self)
-    self._parser.CommentHandler = self.handleComment
+    def __init__(self):
+      """
+        Constructor.
+        @ In, None
+        @ Out, None
+      """
+      ET.XMLParser.__init__(self)
+      self._parser.CommentHandler = self.handleComment
 
-  def handleComment(self,data):
-    """
-      Constructor.
-      @ In, data, string object to parse into comment
-      @ Out, None
-    """
-    self._target.start(ET.Comment,{})
-    self._target.data(data)
-    self._target.end(ET.Comment)
+    def handleComment(self,data):
+      """
+        Constructor.
+        @ In, data, string object to parse into comment
+        @ Out, None
+      """
+      self._target.start(ET.Comment,{})
+      self._target.data(data)
+      self._target.end(ET.Comment)
 
-#set up a parser for this module
+  #set up a parser for this module
 parser = XMLCommentParser()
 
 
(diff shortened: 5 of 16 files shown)


framework/utils/TreeStructure.py
```diff
@@ -321,30 +321,34 @@ def addChild(gNode,tNode,depth=0):
...
-      @ Out, None
-    """
-    self._target.start(ET.Comment,{})
-    self._target.data(data)
-    self._target.end(ET.Comment)
-
-#set up a parser for this module
+if sys.version_info.major > 2:
+  #In Python3, XMLParser is a C implementation, and can't be modified.
+  XMLCommentParser = ET.XMLParser
```
alfoa: So, In python3 we are going to loose the comments or the new C implementation preserves them by default?

joshua-cogliati-inl: Hm, I don't know.  Is there any easy way to test the results?

alfoa: @PaulTalbot-INL Which test uses this class? I do not remember.

joshua-cogliati-inl: Maybe framework.getpot_with_externalXML?

joshua-cogliati-inl: framework/utils.inputParsing will fail in Python 2 if I force it to use XMLParser instead of XMLCommentParser.

joshua-cogliati-inl: I had to update inputParsing to work with python3, and here is the result:
```
framework/utils.inputParsing: Working Directory: /home/jjc/raven/submodules/raven/tests/framework/utils
framework/utils.inputParsing: Running command: python3 testParse.py
framework/utils.inputParsing: Testing XML to XML ...
framework/utils.inputParsing: al '  <DeepOne do="D1">\n' != bl '  <!--comment about DeepOne-->\n'
framework/utils.inputParsing: al '    <DeepTwo do="D2">\n' != bl '  <DeepOne do="D1">\n'
framework/utils.inputParsing: al '      <DeepThree do="D3">\n' != bl '    <!--comment about DeepTwo-->\n'
framework/utils.inputParsing: al '        <DeepFour do="D4">deepest</DeepFour>\n' != bl '    <DeepTwo do="D2">\n'
framework/utils.inputParsing: al '      </DeepThree>\n' != bl '      <DeepThree do="D3">\n'
framework/utils.inputParsing: al '    </DeepTwo>\n' != bl '        <DeepFour do="D4">deepest</DeepFour>\n'
framework/utils.inputParsing: al '    <NotSoDeep do="NSD" />\n' != bl '      </DeepThree>\n'
framework/utils.inputParsing: al '  </DeepOne>\n' != bl '    </DeepTwo>\n'
framework/utils.inputParsing: al '\n' != bl '    <NotSoDeep do="NSD" />\n'
framework/utils.inputParsing: al '  <Objects>\n' != bl '  </DeepOne>\n'
framework/utils.inputParsing: al '    <Object name="MyObj" type="OtherObj">\n' != bl '\n'
framework/utils.inputParsing: al '      <params>1,2,3</params>\n' != bl '  <!--generic comment-->\n'
framework/utils.inputParsing: al '      <alpha>a,b,c</alpha>\n' != bl '  <Objects>\n'
framework/utils.inputParsing: al '    </Object>\n' != bl '    <Object name="MyObj" type="OtherObj">\n'
framework/utils.inputParsing: al '    <Object name="MyObj2" type="OtherObj">\n' != bl '      <params>1,2,3</params>\n'
framework/utils.inputParsing: al '      <params>7,8,9</params>\n' != bl '      <alpha>a,b,c</alpha>\n'
framework/utils.inputParsing: al '      <alpha>x,y,z</alpha>\n' != bl '    </Object>\n'
framework/utils.inputParsing: al '    </Object>\n' != bl '    <Object name="MyObj2" type="OtherObj">\n'
framework/utils.inputParsing: al '  </Objects>\n' != bl '      <params>7,8,9</params>\n'
framework/utils.inputParsing: al '</MyRoot>\n' != bl '      <alpha>x,y,z</alpha>\n'
framework/utils.inputParsing:   ... failures in XML to XML:
framework/utils.inputParsing:       line 5 is not the same!
framework/utils.inputParsing: Testing GetPot to XML ...
framework/utils.inputParsing:   ... passed!
framework/utils.inputParsing: Testing XML to GetPot ...
framework/utils.inputParsing: al '  [./DeepOne]\n' != bl '  [./DeepOne] #comment about DeepOne\n'
framework/utils.inputParsing: al '    [./DeepTwo]\n' != bl '    [./DeepTwo] #comment about DeepTwo\n'
framework/utils.inputParsing: al '  [./Objects]\n' != bl '  [./Objects] #generic comment\n'
framework/utils.inputParsing:   ... failures in GetPot to XML:
framework/utils.inputParsing:       line 12 is not the same!
framework/utils.inputParsing: Testing GetPot to GetPot ...
framework/utils.inputParsing:   ... passed!
framework/utils.inputParsing: Results: ['passed: 2', 'failed: 2']
framework/utils.inputParsing: 
```

PaulTalbot-INL: I don't know that we had a test that explicitly tested retaining comments.  I think we've discussed before focusing on making our own comment tag instead of retaining the default XML comments?

joshua-cogliati-inl: Yes, we discussed that in https://hpcgitlab.inl.gov/idaholab/raven/issues/827 but that doesn't seem to have moved to github.

joshua-cogliati-inl: Added part of that issue to #199

joshua-cogliati-inl: Possible solution for Python3:
https://stackoverflow.com/questions/33573807/faithfully-preserve-comments-in-parsed-
xml-python-2-7

crow/setup3.py
```diff
@@ -18,15 +18,15 @@
 except:
   eigen_flags = ""
 include_dirs=['include/distributions','include/utilities','contrib/include']
-if eigen_flags.startswith("-I"):
+if eigen_flags.startswith(b"-I"):
```
maljovec: This `b` ensures that the string following is encoded in bytes and contains only ASCII characters, whereas python 3 defaults to unicode for strings.

I was going to ask why this was needed here, but after reading the [documentation](https://docs.python.org/2/library/subprocess.html#subprocess.check_output), I realize it is necessary since subprocess.check_output returns a byte string.  The remaining question is if we hit the except block and store an empty unicode `str` type into eigen_flags, does this line complain or evaluate to False as expected?

framework/CustomCommandExecuter.py
```diff
@@ -43,4 +43,4 @@ def execCommand(commandString,self=None,object=None):
     @ In, object, instance, optional, object instance
     @ Out, None
   """
-  execCommandReturn(commandString,self,object)
+  exec(commandString)
```
maljovec: Does this mean `execCommandReturn` can be deleted, or is it called directly elsewhere?

crow/setup3.py
joshua-cogliati-inl: The next line would probably fail if the if statement was true.  It probably should assume that eigen_flags is ascii and decode the bytes to unicode.

framework/CustomCommandExecuter.py
joshua-cogliati-inl: Well, it does have the side effect of putting self and object into the namespace before exec is called.

framework/DataObjects/Data.py
```diff
@@ -639,7 +640,7 @@ def getParam(self,typeVar,keyword,nodeId=None,serialize=False):
     else:
       if typeVar.lower() in ['input','inputs','unstructuredInput']:
         returnDict = {}
-        if keyword in self._dataContainer['inputs'].keys() + self._dataContainer['unstructuredInputs'].keys():
+        if keyword in itertools.chain(self._dataContainer['inputs'].keys(),self._dataContainer['unstructuredInputs'].keys()):
```
maljovec: This makes sense as you don't have to construct the new object that mashes both lists together, also timing shows improvement:

```python
>>> import itertools, timeit
>>> timeit.timeit('[x for x in A+B]', setup = 'A = list(range(1000000)); B = list(range(1000000,2000000))', number=10)
0.8271918296813965
>>> timeit.timeit('[x for x in itertools.chain(A,B)]', setup = 'A = list(range(1000000)); B = list(range(1000000,2000000))', number=10)
0.6930079460144043
```

framework/Samplers/Sampler.py
```diff
@@ -295,7 +295,7 @@ def _readMoreXMLbase(self,xmlNode):
           self.raiseAnError(IOError,'The following are duplicated variables indices listed in the manifestVariablesIndex: ' + str(dups))
         listElement = self.distributions2variablesMapping[dist]
         for var in listElement:
-          self.variables2distributionsMapping[var.keys()[0]]['totDim'] = maxDim #reset the totDim to reflect the totDim of original input space
+          self.variables2distributionsMapping[utils.first(var.keys())]['totDim'] = maxDim #reset the totDim to reflect the totDim of original input space
```
maljovec: Seems safer and also I like the fact that there is not a literal value sitting in the code. Something just feels better about this.

framework/utils/xmlUtils.py
```diff
@@ -206,16 +206,16 @@ def fixXmlText(msg):
     @ Out, msg, string, fixed string
   """
   #if not a string, pass it back through
-  if not isinstance(msg,basestring):
+  if not type(msg).__name__ in ['unicode','str']:
```
maljovec: Can this just use the function you defined in `utils` above?

framework/utils/xmlUtils.py
```diff
@@ -226,7 +226,7 @@ def fixXmlTag(msg):
...
```
maljovec: Again, can this be:

```python
if not utils.isString(msg):
```

setup3.py
```diff
@@ -13,6 +13,22 @@
 # limitations under the License.
 from distutils.core import setup, Extension
 from distutils.command.build import build
+import os
+
+# Replicating the methods used in the RAVEN Makefile to find CROW_DIR,
+# If the Makefile changes to be more robust, so should this
+# We should be doing a search for CROW, I would think, we should not force a
+# directory structure
```
maljovec: Good point.

framework/utils/xmlUtils.py
joshua-cogliati-inl: Yes, if we fix this failure: https://www.moosebuild.org/job/90473/

crow/setup3.py
```diff
@@ -18,15 +18,15 @@
 except:
   eigen_flags = ""
 include_dirs=['include/distributions','include/utilities','contrib/include']
```
maljovec: Just noticed this, but should this line be using `os.path.join`?

crow/setup3.py
```diff
@@ -18,15 +18,15 @@
...
-if eigen_flags.startswith("-I"):
+if eigen_flags.startswith(b"-I"):
   include_dirs.append(eigen_flags[2:].rstrip())
 swig_opts=['-c++','-py3','-Iinclude/distributions','-Iinclude/utilities']
```
maljovec: Same here, I am wondering if these includes should use `os.path.join`, also this is a subset of the includes above, I am wondering if it is at all beneficial to abstract these into variables? This may be overkill, and it is probably better to be explicit about what includes are needed where rather than just throwing all of the includes everywhere.

crow/setup3.py
```diff
@@ -18,15 +18,15 @@
...
-                   Extension('_interpolationND'+ext,['crow_modules/interpolationND'+ext+'.i','src/utilities/ND_Interpolation_Functions.C','src/utilities/NDspline.C','src/utilities/microSphere.C','src/utilities/inverseDistanceWeigthing.C','src/utilities/MDreader.C','src/distributions/randomClass.C'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args)],
+      ext_modules=[Extension('_distribution1D'+ext,['crow_modules/distribution1D'+ext+'.i','src/distributions/distribution.cxx','src/utilities/MDreader.cxx','src/utilities/inverseDistanceWeigthing.cxx','src/utilities/microSphere.cxx','src/utilities/NDspline.cxx','src/utilities/ND_Interpolation_Functions.cxx','src/distributions/distributionNDBase.cxx','src/distributions/distributionNDNormal.cxx','src/distributions/distributionFunctions.cxx','src/distributions/DistributionContainer.cxx','src/distributions/distribution_1D.cxx','src/distributions/randomClass.cxx','src/distributions/distributionNDCartesianSpline.cxx'],include_dirs=include_dirs,swig_opts=swig_opts,extra_compile_args=extra_compile_args),
```
maljovec: I guess as long as I am being picky about the relative paths, this line uses explicit `/` as well.

maljovec: These may be unnecessary fixes, @joshua-cogliati-inl, I will let you comment on whether these should changed.

framework/utils/xmlUtils.py
maljovec: Seems strange, they are in the same directory? How is xmlUtils found but not utils? How about try:

```python
from .utils.utils import isString
```

This way it should know to look for the utils submodule in the current module.

maljovec: This is at least how I got around the issue when breaking the classes into their own submodules.  See, for instance in `PointSet.py`:

```python
from .Data import Data, NotConsistentData, ConstructError
```

crow/setup3.py
joshua-cogliati-inl: Possibly, but since windows supports "/" as a directory separator unless we want to support VMS or classic Mac OS or something else exotic, we are safe doing this.

framework/utils/xmlUtils.py
joshua-cogliati-inl: I did try .utils, and that did not solve the problem when running from make_docs.sh  I don't think it would be hard to solve, I just only tried the first thing I could think of.

crow/setup3.py
alfoa: I would say...better safe than sorry :D

joshua-cogliati-inl: I will make the comment that this same section is in the python2 version setup.py :)

framework/utils/xmlUtils.py
maljovec: Hmm... back to the drawing board. I am going to clone your fork and play around with it briefly. I don't intend to commit anything.

maljovec: @joshua-cogliati-inl, it looks like the problem is how `WriteTex.py` was importing the module. I propose two changes to ensure this works:

1. Add the following line to `framework/utils/__init.py__`
```python
from __future__ import absolute_import
```
2. Change the following line in `doc/user_guide/writeTex.py`:
```python
- sys.path.append(os.path.join(os.getcwd(),'..','..','framework','utils'))
- import xmlUtils
+ sys.path.append(os.path.join(os.getcwd(),'..','..','framework'))
+ from utils import xmlUtils
```

maljovec: The issue seems to be that adding the `framework/utils` directory to the path does not let python make itself aware that the `utils` is a module, thus directly importing `xmlUtils` makes it so that it is not aware of siblings within its own parent module, I guess?  By backing out one directory, python can be made aware that there is a larger module called `utils`, and can find `.utils` beside `xmlUtils` in the module.

maljovec: Lastly, this does not work unless the utils module has the `absolute_import`.

crow/setup3.py
joshua-cogliati-inl: The line above should probably be changed to:
```
eigen_flags = subprocess.check_output(["./scripts/find_eigen.py"]).decode("ascii")
```
so that eigen_flags is ascii.

maljovec:
The solution in the stackoverflow post regarding the XML comment parser passes the offending test case when I use both Python 2 and Python 3 to test it. I have committed this change to this fork.

joshua-cogliati-inl:
FYI: with 
```./run_tests --max-fails 500 ```
the result is:
103 passed, 54 skipped, 0 pending, 311 FAILED

maljovec:
@joshua-cogliati-inl, I am not seeing these failures.  CIVET is reporting positive test results (https://www.moosebuild.org/pr/3519/), and I don't see any errors when I run your branch locally on my machine.  Can you elaborate?

joshua-cogliati-inl:
I edited the comment to make it clear that the failures is with Python3.  Basically, we went from failing during compiling to 103 passed.

maljovec:
I see. Thanks for the clarification.

PR #227, opened 2017-06-22
Closes #226

### LINKED ISSUE #226: timing performance of ensemble models
--------
Issue Description
--------
A constant timing performance as function of the number of samples (in MC)

Performance degrades after 5K samples

##### Please attach the input file(s) that generate this error.  The simpler the input, the faster we can find the issue.

----------------
----------------
- [x] 1. Is it tagged with a type: defect or improvement? ***improvement***
- [x] 2. Is it tagged with a priority: critical, normal or minor? ***critical***
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements? ***n/a***
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.) ***yes***

-------
-------
- [x] 1. If the issue is a defect, is the defect fixed? Not a defect
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.) NA
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)? NA
- [x] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)? NA
- [x] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided? NA

alfoa:
--------
Pull Request Description
--------

#226 

Some unneeded deep copies have been removed and a modification of the reset method in the Data.py has been added

### DIFF
--- a/framework/Models/EnsembleModel.py
+++ b/framework/Models/EnsembleModel.py
@@ -358,7 +358,7 @@ class EnsembleModel(Dummy):
     for key in kwargs["SampledVars"].keys():
       if key in self.modelsDictionary[modelName]['Input']:
         selectedkwargs['SampledVars'][key], selectedkwargs['SampledVarsPb'][key] =  kwargs["SampledVars"][key],  kwargs["SampledVarsPb"][key] if 'SampledVarsPb' in kwargs.keys() else 1.0
-    return copy.deepcopy(selectedkwargs)
+    return selectedkwargs
 
   def createNewInput(self,myInput,samplerType,**kwargs):
     """
@@ -435,7 +435,7 @@ class EnsembleModel(Dummy):
         if targetEvaluations[modelIn].type != 'HistorySet':
           castedUnstructuredInputsValues = {}
           for key in unstructuredInputsValues.keys():
-            castedUnstructuredInputsValues[key] = copy.copy(unstructuredInputsValues[key][-1])
+            castedUnstructuredInputsValues[key] = unstructuredInputsValues[key][-1]
         else:
           castedUnstructuredInputsValues  =  unstructuredInputsValues.values()[-1]
         inputsValues.update(castedUnstructuredInputsValues)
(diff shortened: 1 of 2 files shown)


alfoa:
@mandd

mandd:
Do we need to operate on the function checkTypeRecursively in utils.py?

framework/Models/EnsembleModel.py
```diff
@@ -656,7 +656,7 @@ def _externalRun(self,inRun, jobHandler):
           while not moveOn:
             if jobHandler.availability() > 0:
               # run the model
-              self.modelsDictionary[modelIn]['Instance'].submit(originalInput[modelIn], samplerType, jobHandler, **copy.deepcopy(inputKwargs[modelIn]))
+              self.modelsDictionary[modelIn]['Instance'].submit(originalInput[modelIn], samplerType, jobHandler, **inputKwargs[modelIn])
```
maljovec: I believe I added this deepcopy, if it passes the tests without it, then this should be used instead. I may have been overly cautious when passing things across jobs/threads. This may also be able to be changed on the Steps.py file.  I will have to check that outside of this PR.

maljovec:
@mandd, I did not see much improvement from modifying that to be a string.  I think we can hold off on that for now.

alfoa:
The reason why I did not operate on the "checkTypeRecursively" function is that in reality it should be called only once for each metadata. The problem was that the "reset" method was not keeping the mapping of the metadata type and the checkTypeRecursively was always called. See the method "__getMetadataType" in Data.py

alfoa:
Regarding the profiling. 
With this PR changes the computational times dropped as follows:
- 1000 samples of the PlantModel dropped from 45seconds to 21seconds. 
- 2000 samples of the PlantModel dropped from 92seconds to 49seconds.

PR #229, opened 2017-06-22
Optimizer Preconditioning (Multilevel)

### LINKED ISSUE #228: Preconditioning in Multilevel Optimization
Users from the HYBRID project have requested the ability to precondition lower subspaces when steps are taken in higher subspaces, during multilevel optimization runs.

----------------
----------------

-------
-------
- [ ] 1. If the issue is a defect, is the defect fixed?
- [ ] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [ ] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [ ] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)?
- [ ] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided?

PaulTalbot-INL:
--------
Pull Request Description
--------
Closes #228.

Introduces "preconditioning" to multilevel optimization calculations, wherein inner loops can have an initial guess via an External Model based on the outer-loop subspace variable values.

----------------
----------------
- [x] 1. Review all computer code. *Done by @maljovec  and @joshua-cogliati-inl*
- [x] 2. If any changes occur to the input syntax, there must be an accompanying change to the user manual and xsd schema. If the input syntax change deprecates existing input files, a conversion script needs to be added (see Conversion Scripts). *Manual and xsd updated*
- [x] 3. Make sure the Python code and commenting standards are respected (camelBack, etc.) - See on the [wiki](https://github.com/idaholab/raven/wiki/RAVEN-Code-Standards#python) for details. *Yes*
- [x] 4. Automated Tests should pass, including run_tests, pylint, manual building and xsd tests. If there are changes to Simulation.py or JobHandler.py the qsub tests must pass. *Yes*
- [x] 5. If significant functionality is added, there must  be tests added to check this. Tests should cover all possible options.  Multiple short tests are preferred over one large test. If new development on the internal JobHandler parallel system is performed, a cluster test must be added setting, in <RunInfo> XML block, the node ```<internalParallel>``` to True. *Tests added*
- [x] 6. If the change modifies or adds a requirement or a requirement based test case, the Change Control Board's Chair or designee also needs to approve the change.  The requirements and the requirements test shall be in sync.  *Seconded by @joshua-cogliati-inl*
- [x] 7. The merge request must reference an issue.  If the issue is closed, the issue close checklist shall be done. *Done*
- [x] 8. If an analytic test is changed/added is the the analytic documentation updated/added? *No changes*


tests/framework/AnalyticModels/optimizing/maljovec_precond.py
```diff
@@ -0,0 +1,5 @@
+def evaluate(y):
```
maljovec: Likewise, this preconditioner should be named similar to above, or maybe even more generically since it is just the identity function. Possibly identity_precond?  This could be useful for any number of appropriately aligned test problems.

tests/framework/AnalyticModels/optimizing/plot_functions.py
```diff
@@ -43,9 +43,19 @@ def plotFunction(title,method,low,high,cscale=None,log=True):
   else:
     norm = colors.Normalize()
   ax.pcolormesh(X,Y,Z,norm=norm)
+  #ax.plot([-4.5,4.5],[-4.5,4.5],'k:')
+  #ax.contour(X,Y,Z,colors='k',levels=[1e-5,1e-1,5e-1,1e0,5e0,1e1])
   plt.title(title)
   return X,Y,Z
 
+from maljovech import evaluate as malj
```
maljovec: Does this work? The "h" at the end doesn't cause it to fail to import?

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -265,86 +267,124 @@ def checkConvergence(self):
...
     return convergence
 
-  def _updateConvergenceVector(self, traj, varsUpdate, currentLossValue):
+  def _updateConvergenceVector(self, traj, varsUpdate, currentLossVal):
     """
       Local method to update convergence vector.
       @ In, traj, int, identifier of the trajector to update
       @ In, varsUpdate, int, current variables update iteration number
-      @ In, currentLossValue, float, current loss function value
+      @ In, currentLossVal, float, current loss function value
```
PaulTalbot-INL: This was just to be self-consistent with other variable names, like "oldVal".  It was less work to change this one than the other ones, if I recall correctly.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -265,86 +267,124 @@ def checkConvergence(self):
...
-        oldVal = objectiveOutputs.mean() # TODO should this be from counter[recentOptPoints][traj]?
-        # see if new point is better than old point
-        newerIsBetter = self.checkIfBetter(currentLossValue,oldVal)
-        varK = self.denormalizeData(self.optVarsHist[traj][self.counter['varsUpdate'][traj]])
-        ## convergence values
-        # gradient norm
-        gradNorm  = self.counter['gradNormHistory'][traj][0]
+    # if trajectory is already converged, no need to update. #FIXME then why was this method called?
+    if self.convergeTraj[traj]:
+      return
```
PaulTalbot-INL: I'm running the coverage on this branch to check that.  I'd be surprised if it does, but I wasn't confident enough to remove it during development.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -265,86 +267,124 @@ def checkConvergence(self):
...
+      self.raiseAMessage('Setting convergence for Trajectory "{}" to "False" because of preconditioning.'.format(traj))
+      converged = False
+    else:
+      self.raiseAMessage('Checking convergence for Trajectory "{}":'.format(traj))
+      self.convergenceProgress[traj] = {} # tracks progress for grad norm, abs, rel tolerances
+      converged = False                   # updated for each individual criterion using "or" (pass one, pass all)
+      #printing utility
+      printString = '    {:<21}: {:<5}'
+      printVals = printString + ' (check: {:>+9.2e} < {:>+9.2e}, diff: {:>9.2e})'
+      def printProgress(name,boolCheck,test,gold):
```
PaulTalbot-INL: I don't like where it is, either.  I can make it a module-level method, that feels better to me.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -265,86 +267,124 @@ def checkConvergence(self):
...
+      except KeyError:
+        #we reset the step size, so we don't have a value anymore
+        lastStep = np.nan
+        minStepSizeCheck = False
+      printProgress('Min step size',minStepSizeCheck,lastStep,self.minStepSize)
+      converged = converged or minStepSizeCheck
+
+      # gradient norm
+      if len(self.counter['gradientHistory'][traj][0]) > 0: # that is, if we have gradient evaluations available ...
+        gradNorm = self.counter['gradNormHistory'][traj][0] # TODO the check used to be for len(...[1])>0
```
PaulTalbot-INL: My personal code of comment adding has generally been to include comments before if they are 1) multiline or 2) apply to several following lines of of code, and inline if it applies only to that line.  This is mostly to increase readability by allowing more consecutive lines of code to be adjacent without comments between them.  This might be worth bringing up as a standards question.

The comment in question is mostly a note to self to remind myself how it behaved before the rework of the gradient evaluation location.  I can remove the comment now.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -265,86 +267,124 @@ def checkConvergence(self):
...
+      converged = converged or gradientNormCheck
+
+      # if accepting new point, then "same coordinate" and "abs" and "rel" checks are also valid reasons to converge
+      if newerIsBetter:
+        #absolute tolerance
+        absLossDiff = abs(currentLossVal-oldLossVal)
+        self.convergenceProgress[traj]['abs'] = absLossDiff
+        absTolCheck = absLossDiff <= self.absConvergenceTol
+        printProgress('Absolute Loss Diff',absTolCheck,absLossDiff,self.absConvergenceTol)
+        converged = converged or absTolCheck
```
PaulTalbot-INL: The checks are quite inexpensive, and we offer them as outputs on each iteration to the user, so it is worth checking them all, in my opinion.

Originally, the "or" check was done all in one line, but this was not conducive to future development involving addition of other convergence criteria. In the way it is now, a developer need only copy one of the existing check blocks.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -438,17 +478,18 @@ def localFinalizeActualSampling(self,jobObject,model,myInput):
     # for some reason, Ensemble Model doesn't preserve this information, so wrap this debug in a try:
     try:
       prefix = jobObject.getMetadata()['prefix']
-      self.raiseADebug('Collected sample "{}"'.format(prefix))
     except TypeError:
       prefix = ''
     self.raiseADebug('Collected sample "{}"'.format(prefix))
 
     # TODO REWORK move this whole piece to Optimizer base class as much as possible
+    # TODO what if there's no solution export?  Our "status" and etc fail badly!
```
PaulTalbot-INL: I was too afraid to test this.  I will copy an existing test case and see what happens.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -438,17 +478,18 @@ def localFinalizeActualSampling(self,jobObject,model,myInput):
...
     if self.solutionExport != None and len(self.mdlEvalHist) > 0:
       for traj in self.optTraj:
-        while self.counter['solutionUpdate'][traj] <= self.counter['varsUpdate'][traj]:
+        #does this need to be a while loop?  We can't get behind by 2 points...
+        #while self.counter['solutionUpdate'][traj] <= self.counter['varsUpdate'][traj]:
```
PaulTalbot-INL: I can remove this.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -605,3 +646,13 @@ def getQueuedPoint(self,traj,denorm=True):
...
     return prefix,point
 
+  def proposeNewPoint(self,traj,point):
+    """
+      See base class.  Used to set next recommended point to use for algorithm, overriding the gradient descent.
+      @ In, traj, int, trajectory who gets proposed point
+      @ In, point, dict, input space as dictionary {var:val}
+      @ Out, None
+    """
+    Optimizer.proposeNewPoint(self,traj,point)
```
PaulTalbot-INL: The "propose new point" is currently only used in Preconditioning, where instead of evaluating the gradient to determine the next point, we simply accept the preconditioner's suggestion as a theoretical new point and evaluate the model at the preconditioner-chosen point.  

There isn't any gradient evaluation being performed in this case, as can be seen in the TargetEvaluation output, where the runs with even prefixes are missing on preconditioned points.

framework/Optimizers/Optimizer.py
```diff
@@ -571,12 +596,40 @@ def updateMultilevelDepth(self, traj, depth, optPoint, setAll=False):
...
+      self.raiseADebug('Preconditioning subsets below',oldDepth,range(oldDepth+1,depth+1))
+      #apply changes all the way down
+      for d in range(oldDepth+1,depth+1):
+        precondBatch = self.mlSequence[d]
+        precond = self.mlPreconditioners.get(precondBatch,None)
+        if precond is not None:
+          self.raiseADebug('Running preconditioner on batch "{}"'.format(precondBatch))
+          infoDict = {'SampledVars':self.denormalizeData(optPoint)}
+          _,(results,_) = precond.evaluateSample([infoDict['SampledVars']],'Optimizer',infoDict)
+          # flatten results #TODO breaks for multi-entry arrays
```
PaulTalbot-INL: It turns out there is no possibility to have multi-entry arrays currently, except possibly through a complicated ensemble model operation.  I like the comment as a leg  up in the event future development runs into this limitation.

framework/Optimizers/Optimizer.py
```diff
@@ -571,12 +596,40 @@ def updateMultilevelDepth(self, traj, depth, optPoint, setAll=False):
...
+  def proposeNewPoint(self,traj,point):
+    """
+      Sets a proposed point for the next in the opt chain.  Recommended to be overwritten in subclasses.
+      @ In, traj, int, trajectory who is getting proposed point
+      @ In, point, dict, new input space point as {var:val}
+      @ Out, None
+    """
+    point = copy.deepcopy(point)
+    self.optVarsHist[traj][self.counter['varsUpdate'][traj]] = point
+    self.recommendedOptPoint[traj] = point
```
PaulTalbot-INL: Correct.  Neither entity was designed to change "point", but if either do, it should change in both.

tests/framework/AnalyticModels/optimizing/maljovec.py
```diff
@@ -0,0 +1,17 @@
+# Function designed by D. Maljovech to have sinusoidal minimum
```
PaulTalbot-INL: I see I failed the surname spelling too.  I can change it to anything, no problem.

tests/framework/AnalyticModels/optimizing/maljovec_precond.py
PaulTalbot-INL: I like it named specifically as the preconditioner for a particular function, or not as a preconditioner at all if it has broader application.  I'll change it.

tests/framework/AnalyticModels/optimizing/plot_functions.py
PaulTalbot-INL: I'll look into it.

framework/Optimizers/GradientBasedOptimizer.py
PaulTalbot-INL: Ah, the difference is we call "self.raiseAMessage" within the method, implicitly including "self" in the namespace.  I can add "caller" as a parameter and make it a modular function, but that seems like stretching things a little bit.  Opinions?

framework/Optimizers/GradientBasedOptimizer.py
PaulTalbot-INL: It fails badly.  I'm going to make SolutionExport required for now.

To be honest, you get nothing useful out of the optimizer if you don't use a solution export, so I can't think of a case where you shouldn't have it anyway.

joshua-cogliati-inl: 'SolutionExport required' I assume you mean you are now checking when the input file is read?

framework/Optimizers/GradientBasedOptimizer.py
PaulTalbot-INL: It does not.  Removing.

framework/Optimizers/GradientBasedOptimizer.py
PaulTalbot-INL: Unfortunately not, since it doesn't get assembled then, but once it is assembled ("initialize" phase) I check its existence.

framework/Optimizers/Optimizer.py
maljovec: What is a "multi-entry array"?

PaulTalbot-INL: The objects returned currently are always single-valued, as [1.234].  Multi-entry I intend to convey arrays with more than 1 entry.

framework/Optimizers/GradientBasedOptimizer.py
joshua-cogliati-inl: I think if we can come up with a use of this somewhere else, then it might be worth moving to utils.  Otherwise, I think a one line function can stay local.

framework/Optimizers/GradientBasedOptimizer.py
```diff
@@ -471,17 +505,17 @@ def localFinalizeActualSampling(self,jobObject,model,myInput):
...
-                  self.counter['recentOptHist'][traj][1] = copy.deepcopy(self.counter['recentOptHist'][traj][0])
-                except KeyError:
-                  # this means we don't have an entry for this trajectory yet, so don't copy anything
-                  pass
-                self.counter['recentOptHist'][traj][0] = {'inputs':self.optVarsHist[traj][self.counter['varsUpdate'][traj]],
-                                                          'output':currentObjectiveValue}
               # update status to submitting grad eval points
               self.status[traj]['process'] = 'submitting grad eval points'
+            # if rejecting bad point, keep the old point as the new point; otherwise, add the new one
+            if self.status[traj]['reason'] != 'rejecting bad opt point':
```
joshua-cogliati-inl: Using comparison to a string to determine if the bad point was found, is not the best practice (somewhat expensive computationally, and easy to change in one place and forget to change in another.)

framework/Optimizers/Optimizer.py
```diff
@@ -400,11 +412,19 @@ def initialize(self,externalSeeding=None,solutionExport=None):
...
+      cls,typ,name,model = entry
+      if cls != 'Models' or typ != 'ExternalModel':
+        self.raiseAnError(IOError,'Currently only "ExternalModel" models can be used as preconditioners! Got "{}.{}" for "{}".'.format(cls,typ,name))
+      self.preconditioners[name] = model
+
     self.mdlEvalHist = self.assemblerDict['TargetEvaluation'][0][3]
     self.objSearchingROM = SupervisedLearning.returnInstance('SciKitLearn', self, **{'SKLtype':'neighbors|KNeighborsRegressor', 'Features':','.join(list(self.fullOptVars)), 'Target':self.objVar, 'n_neighbors':1,'weights':'distance'})
     self.solutionExport = solutionExport
+    if self.solutionExport is None:
+      self.raiseAnError(IOError,'The results of optimization cannot be obtained with a SolutionExport in the Step!')
```
joshua-cogliati-inl: I am not sure that this error makes sense (as in it seems to say you should not have a SolutionExport, but is checking that there is no solutionExport.)  Maybe I am misunderstanding.

framework/Optimizers/GradientBasedOptimizer.py
PaulTalbot-INL: Yes, Dan and I went a couple rounds with this trying to determine the best way to implement this.  We thought using an `enum` would work well, but even in python3 enums are slower than comparison to strings OR integers.  I think the findings are in the comments to the merge request #218.

While integer comparison IS a little bit faster than string comparison, it isn't nearly as human-friendly, which is the majority of the reason the "status" was implemented in the previous merge request.

It is definitely possible for me to revert the change, but I wasn't convinced by @maljovec findings that swapping to integers would be worthwhile.  I'm up for being convinced, though.

framework/Optimizers/SPSA.py
```diff
@@ -145,7 +145,7 @@ def localStillReady(self, ready, convergence = False):
         if reason == 'just started':
           self.nextActionNeeded = ('start new trajectory',traj)
           break
-        elif reason in ['seeking new opt point']:
+        elif reason in ['seeking new opt point','received recommended point']:
```
joshua-cogliati-inl: Ditto above comment, integer constants might be better for this than string comparisons.

framework/Optimizers/Optimizer.py
PaulTalbot-INL: I definitely missed the "out" in "without".  Fixing.

framework/Optimizers/SPSA.py
PaulTalbot-INL: In the event it helps, you can find a block describing the process/reason status method in the `__init__` definition of the Optimizer on devel.

framework/Optimizers/GradientBasedOptimizer.py
joshua-cogliati-inl: Okay, that sounds fine.  (I think for string constants, Python actually puts all the identical strings in the module in the same location, and then before comparing the bytes looks and sees if they are the same object first, so I don't think it is much slower than integers.)

PaulTalbot-INL:
When this gets merged (since it is all green!), I would appreciate a "squash and merge".  Thanks!

joshua-cogliati-inl:
This changes the Beale/opt_soln.csv file, which is part of a requirements test.  I am curious why this pull request changes the result?  Thanks.

PaulTalbot-INL:
The introduction of the "gainGrowthFactor" and "gainShrinkFactor" were added as keys (with values) to the Beale case, which resulted in a change in both the input and result.  I did assure the algorithm converges nicely toward the analytic minimum value still, but I'm glad you're checking up on the requirements change.

joshua-cogliati-inl:
Okay, than I approve the requirements test change.  Since both Dan and I approve it, the checklist passes.

PR #409, opened 2017-10-24
Opt, Dev mode for raven_framework

### LINKED ISSUE #408: Speedup through asserts
--------
Issue Description
--------
API checking should not get in the way of fast performance.

Some Python `if` checks dealing with APIs are impossible for the user to hit, and
when called millions of times per run, can slow down RAVEN calculations.

##### Do you have a suggested fix for the development team?
Python has `assert` functionality that provides API checks; however, if python is run with the `-O` option, these assertions are skipped.  We could make use of this to have a `dev` mode and an `opt` mode.

----------------
----------------

-------
-------
- [x] 1. If the issue is a defect, is the defect fixed? __Not a defect__
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.) __N/A__
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?  __email was proposed in PR #409__ 
- [x] 4. If the issue is a defect, does it impact the latest stable branch? If yes, is there any issue tagged with stable (create if needed)? __Not a defect__
- [x] 5. If the issue is being closed without a merge request, has an explanation of why it is being closed been provided? __PR #409__

Issue close checklist is reviewed by Congjian Wang @wangcj05 on Oct. 26, 2017.

PaulTalbot-INL:
--------
Pull Request Description
--------
Closes #408 

#### What are the significant changes in functionality due to this change request?
By default, sets Python to run in `-O` optimized mode, which strips out assertions and creates `.pyo` files.  This allows developers to use `assert` statements within Python code.  Suggestions of when this should be used are detailed [in the wiki](https://github.com/idaholab/raven/wiki/Python-Optimization).

Also adds a "help" option to `raven_framework` that details command-line options.

----------------
----------------
- [x] 4. Automated Tests should pass, including run_tests, pylint, manual building and xsd tests. If there are changes to Simulation.py or JobHandler.py the qsub tests must pass. __All Pass__
- [x] 5. If significant functionality is added, there must  be tests added to check this. Tests should cover all possible options.  Multiple short tests are preferred over one large test. If new development on the internal JobHandler parallel system is performed, a cluster test must be added setting, in <RunInfo> XML block, the node ```<internalParallel>``` to True. __N/A__
- [x] 7. The merge request must reference an issue.  If the issue is closed, the issue close checklist shall be done. #408
- [x] 8. If an analytic test is changed/added is the the analytic documentation updated/added? __N/A__

Checklists were reviewed by Congjian Wang @wangcj05 on Oct. 26, 2017


raven_framework
```diff
@@ -11,5 +11,54 @@ SCRIPT_DIR=`(cd $SCRIPT_DIRNAME; pwd)`
...
+  echo '    Do not attempt to run the model executable, but read from existing output files.'
+  echo ''
+  echo '  interactive'
+  echo '    Run RAVEN in GUI mode if possible.'
+  echo ''
+  echo '  interactiveCheck'
+  echo '    Run tests on interactive GUI.'
+  echo ''
+elif [ "$MODE" = 'opt' ]; then
+  python -O $SCRIPT_DIR/framework/Driver.py "$ARGS"
```
wangcj05: I think it is better to print the description before we run. Something like "echo 'Runs RAVEN in 'OPT' mode'"

raven_framework
```diff
@@ -11,5 +11,54 @@ SCRIPT_DIR=`(cd $SCRIPT_DIRNAME; pwd)`
...
+else
+  python $SCRIPT_DIR/framework/Driver.py "$ARGS"
```
wangcj05: Same as previous comments.

raven_framework
```diff
@@ -11,5 +11,54 @@ SCRIPT_DIR=`(cd $SCRIPT_DIRNAME; pwd)`
...
+    # display "help" print
+    MODE='help'
+    continue
+  fi
+  # otherwise, pass through arguments
+  ARGS[${#ARGS[@]}]="$var"
+done
+
+# run Driver.py based on the mode chosen
+if [ "$MODE" = 'help' ]; then
```
wangcj05: I think we should set $MODE = 'help' if $@ is empty as well

raven_framework
```diff
@@ -11,5 +11,54 @@ SCRIPT_DIR=`(cd $SCRIPT_DIRNAME; pwd)`
...
+# set up run keywords
+# "ARGS" stores command line arguments not treated in this file (passed through)
+declare -a ARGS
+# by default run in "opt" mode
+MODE='opt'
+# loop through arguments and look for those that are handled before Python is called
+for var in "$@"; do
+  if [ "$var" = '-D' ]; then
+    # run in development mode (assertions not stripped)
+    MODE='dev'
```
wangcj05: MODE is set to 'dev' but not referenced in the later code.

PaulTalbot-INL: I turned the "mode" check into a "case" statement and exit on a catch-all, so dev is explicitly used.

raven_framework
PaulTalbot-INL: Done.

raven_framework
PaulTalbot-INL: Done.

raven_framework
PaulTalbot-INL: Done.

raven_framework
```diff
@@ -31,34 +31,49 @@ for var in "$@"; do
...
+    echo ''
+    echo 'Usage:'
+    echo '  raven_framework [-D][interfaceCheck][interactive][interactiveCheck] file'
+    echo ''
+    echo 'Description:'
+    echo '  Runs RAVEN using the input file given by "file".'
+    echo ''
+    echo 'Options:'
+    echo '  -D'
+    echo '    Development mode.  Turns Python "assert" statements on.'
```
wangcj05: add the description about the 'opt' mode

PaulTalbot-INL: I don't think so.  There is no flag or option for it, so it does not belong in the description of "help", I believe.

wangcj05:
@PaulTalbot-INL In addition, an email should be proposed to the raven developers about these changes. Last comment, should we use 'xrange' instead of 'range' in the wiki optimization page?

raven_framework
wangcj05: For me it is more clear to add this description. Because if the user just run 'raven_framework' without the input file, the current 'help' message will guide the user to run the code in 'dev' mode instead of 'opt' mode. I think we should put 'opt' there, and make it default as the code does.

PaulTalbot-INL:
I do include a difference between iterators and lists, which is the difference between xrange and range, but if you want to specifically bring out that option, by all means add a section.

Proposed email:
We have recently modified how RAVEN runs to optimize performance.  In particular, we are now by default using the "-O" parameter in Python, which at this point simply removes assertion checks.  We only use assertions during development; they should not be reachable when running RAVEN as normal.

To summarize, you might see RAVEN speed up a little, but don't need to do or change anything if you're using RAVEN and not developing it.

For developers, passing the "-D" option to "raven_framework" will re-enable the "assert" statements, allowing for more effective debugging when implementing new features in RAVEN.

Thank you and enjoy!

wangcj05:
The email is good to me, maybe you can add "For developers, we recommend to use 'assert' to do the internal check instead of 'if' statement."

PaulTalbot-INL:
Added to the [development guide].(https://github.com/idaholab/raven/wiki/Developer_Information#assertions)

raven_framework
wangcj05: any thoughts here?

PaulTalbot-INL: I don't feel like the help guides the user to run in dev mode, unless they feel like they want to run with assert statements on.  The usage syntax marks the default operation as being without any optional flags.  I'd rather remove the "help" than try and replicate the user manual in it; why do you feel the user is being guided towards dev mode?

PaulTalbot-INL: I increased the verbosity and added a "default usage" mode, which is somewhat unorthodox but might be more clear to users new to *nix.

wangcj05: First, I agree with you we do not want to replicate the user manual in the code. But a brief clear description I think is also useful. Second, I do not think we have documented this new feature in the user manual (we may not need now since this changes only affect the developers), that's why I suggest to put it here. Anyway, my thoughts may be wrong, I really appreciate that you made the changes. To me it is more clear, however it may the have no difference to the users.

raven_framework
```diff
@@ -50,27 +50,36 @@ case $MODE in
...
+    echo '      Do not attempt to run the model executable, but read from existing output files.'
+    echo ''
+    echo '    interactive'
+    echo '      Run RAVEN in GUI mode if possible.'
+    echo ''
+    echo '    interactiveCheck'
+    echo '      Run tests on interactive GUI.'
     echo ''
     ;;
   *)
```
wangcj05: Thanks for the changes. To me it is more clear.

joshua-cogliati-inl:
Just out of curiosity, when run_tests is done, does RAVEN run in debug or opt mode?

wangcj05:
Thanks for the changes, and I think this PR is good to me.

PaulTalbot-INL:
@joshua-cogliati-inl That's a great question.  I don't even know what the right answer is.

If you run tests in dev mode, and you fail an assertion, either 1) the test would have passed, which means the assertion is wrong and should probably be changed; or 2) the test couldn't have passed, in which case you're probably still developing, but maybe you still want to run tests with assertions present to see who fails.

If you run tests in opt mode, well, they will run microscopically faster.

This said, I guess I don't see much reason to run the tests in opt mode.  I'll see if I can change that easily.

wangcj05:
It is good to me. What do you think? @joshua-cogliati-inl

joshua-cogliati-inl:
"the test would have passed, which means the assertion is wrong and should probably be changed" or the assertion caught something that is a subtle problem that is not caught by actual test, as in I have had assertions fail that would result in a wrong number, but not an outright failure.

I don't have time to look at this, but I have no objections to it being merged.

wangcj05:
OK, I will do the checklist, and additional issue #410 is created and will be resolved in the future PR.

wangcj05:
Now this PR is ready to go. Thanks @PaulTalbot-INL, Great Job!

PR #438, opened 2017-11-27
Talbpaul/dummy output

PaulTalbot-INL:
--------
Pull Request Description
--------
Fixes methods `evaluateSample` and `collectOutput` on both `Dummy` and `ExternalModel` so that they work with the new data object API.  `evaluateSample` returns the expected realization, and `collectOutput` is functionally a pass-through to the data object output collection.

Note this is being merged to a rework branch, not devel.


alfoa:
Good WP

PR #441, opened 2017-11-28
Loading dataset fully from dict

PaulTalbot-INL:
--------
Pull Request Description
--------
Allows loading a DataSet directly from a dictionary using `data.load(dict, style='dict',dims={})` syntax.

### DIFF
--- a/framework/DataObjects/XDataSet.py
+++ b/framework/DataObjects/XDataSet.py
@@ -843,13 +843,13 @@ class DataSet(DataObject):
     providedVars = set(source.keys())
     requiredVars = set(self.getVars('input')+self.getVars('output'))
     ## determine what vars are metadata (the "extra" stuff that isn't output or input
+    # TODO don't take "extra", check registered meta explicitly
     extra = list(providedVars - requiredVars)
     self._metavars = extra
     ## figure out who's missing from the IO space
     missing = requiredVars - providedVars
     if len(missing) > 0:
       self.raiseAnError(KeyError,'Variables are missing from "source" that are required for this data object:',missing)
-    # TODO order entries by self._inputs, self._outputs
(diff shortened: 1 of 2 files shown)


framework/DataObjects/XDataSet.py
```diff
@@ -811,6 +818,62 @@ def _fromCSV(self,fname,**kwargs):
...
+        for v,val in enumerate(values):
+          ## coordinates come from each dimension, specific to the "vth" realization
+          coords = dict((dim,source[dim][v]) for dim in varDims)
+          ## swap-in-place the construction; this will likely error if there's inconsistencies
+          values[v] = self.constructNDSample(val,varDims,coords,name=var)
+        #else:
+        #  pass # TODO need to make sure entries are all single entries!
+      data[:,i] = values
+    # set up collector as cached nd array of values
+    self._collector = cached_ndarray.cNDarray(values=data,dtype=object)
```
mandd: why not a numpy standard array?

framework/DataObjects/XDataSet.py
```diff
@@ -216,16 +216,15 @@ def extendExistingEntry(self,rlz):
     # modify outputs to be pivot-dependent
     toRemove = []
     # set up index vars for removal
-    allDims = self.getDimensions()
-    for var in rlz.keys():
+    for var in self.indexes:
```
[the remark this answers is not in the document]
PaulTalbot-INL: Yes, I think so.  Like... `toRemove = (var for var in self.indexes if var in allDims)` ?

framework/DataObjects/XDataSet.py
PaulTalbot-INL: That's a good question.  Given that we call `asDataset()` right afterwords, we can probably get away without using the cached array.

PR #443, opened 2017-11-28
Talbpaul/register metadata

PaulTalbot-INL:
--------
Pull Request Description
--------
Adds a registration system for keeping pointwise metadata.

Every RAVEN entity can call `self.addMetaKeys` to register pointwise metadata keys.  If a variable in a realization isn't an input, output, or registered metadata, the data object will ignore it.


framework/Models/ExternalModel.py
```diff
@@ -250,10 +250,12 @@ def evaluateSample(self, myInput, samplerType, kwargs):
     """
     Input = self.createNewInput(myInput, samplerType, **kwargs)
     inRun = copy.copy(self._manipulateInput(Input[0][0]))
+    print('DEBUG inrun:',inRun)
```
wangcj05: This can be removed

PR #449, opened 2017-11-30
Extend "remove" with "variable" to do collector AND data

PaulTalbot-INL:
calling `data.remove(variable='var')` now removes `var` from both `data._data` and `data._collector`.


alfoa:
I checked the speed of the "potential slow loop" ... it takes ~1e-5 for 3000 variables. I think it is acceptable

PR #452, opened 2017-12-01
check data alignment works

PaulTalbot-INL:
Implements `DataSet.checkDataAlignment` for checking alignment.

### DIFF
--- a/framework/DataObjects/XDataSet.py
+++ b/framework/DataObjects/XDataSet.py
@@ -211,6 +211,39 @@ class DataSet(DataObject):
       self._collector = self._newCollector(width=self._collector.width,dtype=self._collector.values.dtype)
     return self._data
 
+  def checkIndexAlignment(self,indexesToCheck=None):
+    """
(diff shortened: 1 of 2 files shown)


framework/DataObjects/XDataSet.py
```diff
@@ -211,6 +211,39 @@ def asDataset(self):
       self._collector = self._newCollector(width=self._collector.width,dtype=self._collector.values.dtype)
     return self._data
 
+  def checkIndexAlignment(self,indexesToCheck=None):
```
wangcj05: I need this function also able to check the index alignment for given list of variables along the index.

PaulTalbot-INL: We talked in person.

It is not possible currently for a realization to have inconsistent pivot values within a single realization, only between two realizations.

As such, there is no difference between checking alignment as is, or checking after excluding certain variables.  It would not be too hard to implement this, but currently is not useful.

PaulTalbot-INL:
Any other objections, or can we merge this?

Issue #456, opened 2017-12-04
CSVs and Metadata


PaulTalbot-INL:
There is some discussion and decisions that need to be made regarding CSV files, metadata, and restarting.

## How it used to work
Previous to the data object rework, a CSV file was considered both a user-friendly output option that could be loaded up in a spreadsheet tool and looked at, as well as a data storage option that could be restarted from later.

## The problem
With the data object rework, we now include pointwise metadata with the main point as additional columns, instead of in a separate dictionary/XML tree, whereas general metadata is still stored in an XML structure.  This provided immense speedup in IO for data objects.

Printing these CSVs presents a conundrum.

 - If we omit printing the pointwise metadata such as `ProbabilityWeight` entirely, then the CSV cannot be restarted.
 - If we force the user to have the metadata in their CSV, then we remove power from the user to control their output files.
 - If, when printing the CSVs, we put the metadata back into the XML, we notice slowdown in writing.

As of the dataobject rework currently, restarting a `MultiRun` sampling using a CSV will always work, but loading a CSV without the metadata printed will not work in postprocessors like `BasicStatistics`.

## Moving forward
There are a few possibilities for how to move forward.
 1. By default print metadata into the CSVs.  If the user requests specific variables in their CSV, we instead push the metadata into the auxiliary XML file.
 1. Disallow restarting from a CSV except in the case where the user does not request special printing.  Instead, only allow restarting from databases.
 1. Add a third output CSV, an auxiliary CSV, to house the pointwise metadata.  We can either make the auxiliary CSV capable to be disabled by the user or forced regardless.

----------------
----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [x] 1. Is it tagged with a type: defect or improvement?
- [x] 2. Is it tagged with a priority: critical, normal or minor?
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements?
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [ ] 1. If the issue is a defect, is the defect fixed?
- [ ] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
[...]

PR #458, opened 2017-12-04
Sampler Restarts

PaulTalbot-INL:
Gets sampler restarts working in the new dataobject environment.  See #456 for some obstacles coming down the pike that aren't problems yet.


framework/Steps.py
```diff
@@ -570,6 +570,7 @@ def _localInitializeStep(self,inDictionary):
       else:
         self._outputCollectionLambda.append((lambda x: x[1].addOutput(), outIndex))
         self._outputDictCollectionLambda.append((lambda x: x[1].addOutput(), outIndex))
+    self._registerMetadata(inDictionary)
```
wangcj05: This change will break my recent changes in the basic statistics, could you tell me why need to move this?

PaulTalbot-INL: Hm.  I'm guessing it's later in the call, right?

The problem is probably that PostProcessor does not inherit from `BaseClass`, so the methods `provideExpectedMetaKeys` and `addMetaKeys` are missing from the postprocessors.

The short fix is to add those two methods, basically exactly as they are in the Step.

PaulTalbot-INL: We talked in person; the problem was that the sampler was calling `addMetaKeys` in `localGenerateInput` instead of `localInitialize` or `localInputsAndChecks`.

PR #461, opened 2017-12-06
dataobject rework: Optimizer

PaulTalbot-INL:
- Fixes optimizer
 - Fixes models to assure np.atleast_1d on evaluateSample, which prevents samplers etc from doing the same
 - Fixes CSV inf/nan reading
 - Adds clusterLabel as a tool for outstream printing


PaulTalbot-INL:
I'd like to note the vast majority of the files changed are simple formatting cleanups of test inputs.

These occurred because I ran the optimizer conversion script on the tests in the regression system.

framework/Optimizers/Optimizer.py
```diff
@@ -908,7 +907,7 @@ def generateInput(self,model,oldInput):
     self.localGenerateInput(model,oldInput)
     ####   UPDATE STATICS   ####
     # get trajectory asking for eval from LGI variable set
-    traj = self.inputInfo['trajectory']
+    traj = self.inputInfo['trajID'] - 1
```
wangcj05: Sorry, I did not get the point here?

framework/Optimizers/SPSA.py
```diff
@@ -193,7 +193,8 @@ def localStillReady(self, ready, convergence = False):
           #for i in range(1,self.gradDict['numIterForAve']*2,2):
           for i in self.perturbationIndices:
             evalIndex = self._checkModelFinish(traj,self.counter['varsUpdate'][traj],i)[1]
-            outval = self.mdlEvalHist.getParametersValues('outputs',nodeId='ReconstructEnding')[self.objVar][evalIndex]
+            # OLD outval = self.mdlEvalHist.getParametersValues('outputs',nodeId='ReconstructEnding')[self.objVar][evalIndex]
```
wangcj05: remove?

tests/framework/CodeInterfaceTests/generic_parallel.xml
```diff
@@ -6,12 +6,13 @@
     <created>2017-08-24</created>
     <classesTested>Models.Code.GenericCode, Files</classesTested>
     <description>
-        Tests using "Files" in an internalParallel run (which tests pickling and unpickling them).
+        Tests using &quot;Files&quot; in an internalParallel run (which tests pickling and unpickling them).
```
wangcj05: &quot?

framework/Optimizers/Optimizer.py
PaulTalbot-INL: Historically, the "trajID" was stored as "trajectory + 1" for limitations in the old HistorySet.  It's easier to keep that tradition than try to change it everywhere in the optimizer.

framework/Optimizers/SPSA.py
PaulTalbot-INL: Yes, thank you, removing ...

tests/framework/CodeInterfaceTests/generic_parallel.xml
PaulTalbot-INL: It appears that when the conversion script runs, it doesn't write out the quotation marks well.  Interesting.  I'll get rid of them so it doesn't come up again.

alfoa:
I know that you are going to hate me, but can we avoid all those reformatting of the input files ? It is adding hundreds of diffs that are not strictly needed....

PaulTalbot-INL:
I don't know a way to avoid the reformatting.  Any time someone runs a conversion script it would make these same changes.

alfoa:
I understand...but the conversion script has been run on the all test suite when in reality was supposed to be run on the tests that got a change in the syntax only. Or, if the the XML trees are equivalent (the read one and the new one), the test should not be reprented. I.e.
```
  if etree.tostring(h1) == etree.tostring(h2):
    # do not print
```

alfoa:
I added a new option in the conversion script named ```--no-rewrite``` that avoids to rewrite the input file in case the content did not change.

tests/framework/Optimizers/infinite.xml
```diff
@@ -6,7 +6,7 @@
     <created>2017-06-13</created>
     <classesTested>Optimizer</classesTested>
     <description>
-      This test runs the optimization on 1/x, returning "inf" at x=0.
+      This test runs the optimization on 1/x, returning &quot;inf&quot; at x=0.
```
alfoa: &quot

alfoa:
Ok.
I reverted all the tests that are not part of this PR. 
@mandd @wangcj05 @PaulTalbot-INL whenever you use the converScript try to use the ```--no-rewrite``` flag to avoid this. 

Once we merge the new structure in devel, we can create a PR just to prettify the tests

alfoa:
```
framework/Optimizers.Infinite: Traceback (most recent call last):
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Driver.py", line 279, in <module>
framework/Optimizers.Infinite:     raven()
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Driver.py", line 232, in raven
framework/Optimizers.Infinite:     simulation.run()
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Simulation.py", line 821, in run
framework/Optimizers.Infinite:     stepInstance.takeAstep(stepInputDict)
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Steps.py", line 308, in takeAstep
framework/Optimizers.Infinite:     self._initializeStep(inDictionary)
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Steps.py", line 243, in _initializeStep
framework/Optimizers.Infinite:     self._localInitializeStep(inDictionary)
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Steps.py", line 586, in _localInitializeStep
framework/Optimizers.Infinite:     newInput = self._findANewInputToRun(inDictionary[self.samplerType], inDictionary['Model'], inDictionary['Input'], inDictionary['Output'])
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Steps.py", line 702, in _findANewInputToRun
framework/Optimizers.Infinite:     found,newInp = sampler.generateInput(model,inputs)
framework/Optimizers.Infinite:   File "/Users/alfoa/projects/raven_github/raven/framework/Optimizers/Optimizer.py", line 912, in generateInput
framework/Optimizers.Infinite:     self.values.update(self.denormalizeData(self.mlStaticValues[traj]))
framework/Optimizers.Infinite: TypeError: unhashable type: 'numpy.ndarray'
```

alfoa:
I do not understand.... If I run the tests in this branch (```./run_tests --max-fail=10000  -j8```) I get the following:
```118 passed, 60 skipped, 0 pending, 144 FAILED```
If I run the tests in the branch  ```dataobject-rework``` I get the following:
```207 passed, 61 skipped, 0 pending, 268 FAILED```
Did we delete the tests here? or they are not picked up for whatever reason?

**Comment by PaulTalbot-INL** [comment]
I can't imagine there's that many tests removed.  As far as I'm aware, the only removed tests should be the 4 or 5 in tests/framework/Samplers/Restart/Truncated.

File: framework/DataObjects/XDataSet.py [hunk 5]
```diff
@@ -918,11 +918,11 @@ def _fromCSVXML(self,fName):
...
     try:
       meta,_ = xmlUtils.loadToTree(fName+'.xml')
+      self.raiseADebug('Reading metadata from "{}.xml"'.format(fName))
+      haveMeta = True      
     except IOError:
       haveMeta = False
-    finally:
-      self.raiseADebug('Reading metadata from "{}.xml"'.format(fName))
-      haveMeta = True
+
```
PaulTalbot-INL: I don't think this change actually changes anything ...

alfoa: It does...

```finally:``` is always executed and you were setting the ```haveMeta = True``` in there. 
So even if the ```except``` was setting the ```haveMeta = False``` the finally was always resetting it.

alfoa:
In the email I just sent there is the list of the "missing" tests. For example, all the ROM tests are not present but if I run the command ```./run_tests --re=framework/ROM``` the tests are executed.
It looks like there is a problem with the ```run_tests``` script that got ```triggered``` by this PR...
Very strange... can anyone take a look at it?

PaulTalbot-INL:
What was your process for reverting converted tests that had been added before?  I don't see any changes to any "tests" files or run scripts anywhere in this PR.

framework/DataObjects/XDataSet.py
PaulTalbot-INL: Oh, it should definitely be "else" instead of "finally".  True.  We should avoid putting too many things in the "try" though, as that is prone to misguided error statements.

alfoa:
```git checkout **commit ID** -- file1/to/restore file2/to/restore```

PaulTalbot-INL:
I did a `wc` on `get_coverage_tests.py` on both dataobject-rework and talbpaul/optimizer and it reports 490 "words" on either branch.  I'm running the tests now to confirm what you're seeing.

PaulTalbot-INL:
Turns out the bug was the test harness thread pool crashing silently.  The item causing the crash was using np.isnan instead of pd.isnull in unordered CSV checking.  As it turns out, np.isnan fails badly if the argument is not a standard numpy dtype.

alfoa:
we can. Thank you Paul for addressing this.

alfoa:
I think it is good to go now.

PR #474, opened 2017-12-12
added hierarchical methods

PaulTalbot-INL:
Adds two hierarchical methods to the XDataSet, and a unit test to cover them.

Also introduces protected meta key `RAVEN_HierLevel`, although I'm not sure if it's really necessary as it's the same thing as `len(data['prefix'].split('_'))`.

To get the prefixes of realizations that are path endings, call `_getPathEndings`.  To get the paths for any one ending, call `_reconstructPath`.


PaulTalbot-INL:
This is ready for review now, @alfoa.

framework/DataObjects/TestHierarchal.py
```diff
@@ -0,0 +1,409 @@
...
+    @ Out, res, bool, True if same
+  """
+  res = value == expected
+  if update:
+    if res:
+      results["pass"] += 1
+    else:
+      print("checking string",comment,'|',value,"!=",expected)
+      results["fail"] += 1
+  return res
```
alfoa: probably since we are repeating these methods in each unit test, we should collect them in a single module and import them.

PaulTalbot-INL: I agree but it will take passing in "results" and changing all the method calls. It should be done, but it's a bit of a pain.

PR #478, opened 2017-12-13
Doing better than "object" typing in DataObject

PaulTalbot-INL:
Now keeps track of each variable's type and sets it when the first realization is received.

~~Note that integer types may be wiped out if 1) data is collected, then 2) asDataset() is called, then 3) more data is collected, then 4) asDataset() is called again.  In this case they are converted to float types.~~

~~I've got a question about this out there, we'll see if I find a way to preserve integer types.~~


PaulTalbot-INL:
For personal reference, the query is [here](https://stackoverflow.com/questions/47802335/python-xarray-changing-dtypes-when-merging).

PR #494, opened 2017-12-20
Fixing Tests 1

PaulTalbot-INL:
Fixed framework/DataObjects.dataObject_metadata
Fixed framework/DataObjects.load_csv
Fixed framework/DataObjects.load_csv_history
Fixed framework.testGridWithConstants
Fixed framework/PostProcessors/RavenOutputPostProcessor.dynamic
Fixed framework/PostProcessors/RavenOutputPostProcessor.invalidXMLtag
Fixed framework/PostProcessors/RavenOutputPostProcessor.static
Fixed framework/Optimizers.Infinite
Fixed framework/ErrorChecks.badHistory
Fixed framework/ErrorChecks.badPointSet

Also fixes some XSD validation, also generally fixes "constant" variables.


PaulTalbot-INL:
Currently running tests on Falcon.  Will post results as they show up, although they should be the same as the qsub tests.

PaulTalbot-INL:
Okay this is ready now.  Test diff results:
```
*****************
Number of success first file:
352
357
NUMBER OF SUCCESSES INCREASED: +5
**** NUMBER OF TESTS IN THE TWO FILES ARE DIFFERENT: ****
**** first file :536 /= second file: 532****
TOTAL TIME OF TESTS DECREASED       : reference_time(499.5) > new_time(469.3)
AVERAGE TIMES FOR SUCCESS TESTS ARE : reference_time(1.41903409091) v.s. new_time(1.31456582633)
```

PaulTalbot-INL:
Wait, the numbers don't add up.  I fixed more files than it says pass additionally.  Investigating perpetually....

PaulTalbot-INL:
Because I added the restriction we discussed to not use the same data object as both input and output of a MultiRun step, a few tests started failing.

I fixed those.

Additionally, there were some complications in the event the user provides a dict to load from (from, say, and ExternalPostProcessor) and provides an ND array instead of a np.array(dtype=object) of np.arrays.  I made an "if" to cover it, but it's definitely slower than if the typing is correct.  Final test results are pending.

framework/PostProcessors/RavenOutput.py
```diff
@@ -37,218 +37,223 @@ class RavenOutput(PostProcessor):
...
+#      @ In, inPath, string, |-separated list defining path from root (not including root)
+#      @ In, fileName, string, used in error
+#      @ Out, desiredNode, xml.etree.ElementTree.Element, desired node
+#    """
+#    #improve path format
+#    path = '/'.join(c.strip() for c in inpPath.strip().split('/'))
+#    desiredNode = xmlUtils.findPath(root,path)
+#    if desiredNode is None:
+#      self.raiseAnError(RuntimeError,'Did not find "%s/%s" in file "%s"' %(root.tag,path,fileName))
+#    return desiredNode
```
wangcj05: Do we need all this commented lines?

framework/PostProcessors/__init__.py
```diff
@@ -32,7 +32,7 @@
 from .InterfacedPostProcessor import InterfacedPostProcessor
 from .LimitSurface import LimitSurface
 from .LimitSurfaceIntegral import LimitSurfaceIntegral
-from .RavenOutput import RavenOutput
+# from .RavenOutput import RavenOutput # deprecated for now
```
wangcj05: remove?

framework/PostProcessors/__init__.py
```diff
@@ -66,10 +66,11 @@
...
            'LimitSurfaceIntegral',
-           'RavenOutput',
            'SafestPoint',
            'TopologicalDecomposition',
            'DataMining',
            'Metric',
            'CrossValidation',
            'ETImporter'] + additionalModules
+
+#           'RavenOutput', # deprecated for now
```
wangcj05: rm?

framework/Samplers/Sampler.py
```diff
@@ -711,9 +711,7 @@ def generateInput(self,model,oldInput):
       # DO format the data as atleast_1d so it's consistent in the ExternalModel for users (right?)
       rlz['inputs'] = dict((var,np.atleast_1d(inExisting[var])) for var in self.restartData.getVars('input'))
       rlz['outputs'] = dict((var,np.atleast_1d(inExisting[var])) for var in self.restartData.getVars('output'))
-      rlz['metadata'] = {'prefix':self.inputInfo['prefix'],
-                         'ProbabilityWeight':self.inputInfo['ProbabilityWeight'],
-                        }
+      rlz['metadata'] = copy.deepcopy(self.inputInfo) # TODO need deepcopy only because inputInfo is on self
```
wangcj05: why do we need to copy everything here?

tests/framework/Samplers/EnsembledSampler/test_EnsembleSamplerConstants.xml
```diff
@@ -74,6 +77,8 @@
   </Steps>
 
   <DataObjects>
+    <PointSet name='input'>
+      <Input>x2,x1,constantine</Input>
```
wangcj05: The pointset is not closed with ```</PointSet>```

tests/framework/Samplers/Restart/test_restart_csv.xml
```diff
@@ -23,14 +23,14 @@
...
-      <Input class="DataObjects" type="PointSet">solns</Input>
+      <Input class="DataObjects" type="PointSet">restart</Input>
       <Model class="Models" type="ExternalModel">fine</Model>
       <Sampler class="Samplers" type="MonteCarlo">mc</Sampler>
       <Output class="DataObjects" type="PointSet">solns</Output>
     </MultiRun>
     <IOStep name="load">
       <Input class="Files" type="">csv</Input>
-      <Output class="DataObjects" type="PointSet">solns</Output>
+      <Output class="DataObjects" type="PointSet">restart</Output>
```
wangcj05: what will happen here? The ```restart``` is used as input in previous MultiRun, and used as output in this IOStep.

framework/DataObjects/XDataSet.py
```diff
@@ -1037,44 +1052,39 @@ def _fromCSV(self,fileName,**kwargs):
...
+      samples = range(len(df.index))
+      df[self.sampleTag] = samples
     # create arrays from which to create the data set
     arrays = {}
     for var in self.getVars():
       if var in dims.keys():
-        data = panda[[var,self.sampleTag]+dims[var]]
+        data = df[[var,self.sampleTag]+dims[var]]
         data.set_index(self.sampleTag,inplace=True)
         ndat = np.zeros(len(samples),dtype=object)
```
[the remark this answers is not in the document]
PaulTalbot-INL: Not changed in this PR.  This might be suggested as part of the dataobject-rework merge?

framework/DataObjects/XDataSet.py
```diff
@@ -1425,10 +1450,10 @@ def _setDataTypes(self,rlz):
       self.types = [None]*len(self.getVars())
       for v,name in enumerate(self.getVars()):
         val = rlz[name]
-        if isinstance(val,np.ndarray):
-          self.types[v] = self._getCompatibleType(val.item(0))
-        else:
-          self.types[v] = self._getCompatibleType(val)
+        #if isinstance(val,np.ndarray):
+        #  self.types[v] = self._getCompatibleType(val.item(0))
+        #else:
```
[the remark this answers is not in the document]
PaulTalbot-INL: removed

framework/PostProcessors/ExternalPostProcessor.py
```diff
@@ -189,8 +189,8 @@ def collectOutput(self, finishedJob, output):
     if isinstance(output,Files.File):
       self.raiseAWarning('Output type File not yet implemented. I am going to skip it.')
     elif output.type == 'HDF5':
-      self.raiseAWarning('Output type ' + type(output).__name__
-                         + ' not yet implemented. I am going to skip it.')
+      self.raiseAnError(NotImplementedError,'Output type ' + type(output).__name__
+                         + ' not yet implemented!') # I am going to skip it.')
```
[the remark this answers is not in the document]
PaulTalbot-INL: removed

framework/PostProcessors/Factory.py
```diff
@@ -35,7 +35,7 @@
 from .InterfacedPostProcessor import InterfacedPostProcessor
 from .LimitSurface import LimitSurface
 from .LimitSurfaceIntegral import LimitSurfaceIntegral
-from .RavenOutput import RavenOutput
+# from .RavenOutput import RavenOutput # deprecated for now
```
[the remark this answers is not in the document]
PaulTalbot-INL: I don't think so, since we talked about a possible future use for the RavenOutput.  I think it's a good placeholder.

framework/PostProcessors/RavenOutput.py
PaulTalbot-INL: Yes, otherwise it counts against us in the coverage check; however, the code is likely to be very useful, as it still applies to the XML form of the general metadata.

framework/PostProcessors/__init__.py
PaulTalbot-INL: same

framework/PostProcessors/__init__.py
PaulTalbot-INL: same

framework/Samplers/Sampler.py
PaulTalbot-INL: How do you know what to copy?  This allows each sampler to do as it needs to, which it turns out is required because "prefix" and "probabilityweight" are not generally sufficient.

tests/framework/Samplers/EnsembledSampler/test_EnsembleSamplerConstants.xml
PaulTalbot-INL: I wonder why this didn't come up in the testing.....

tests/framework/Samplers/Restart/test_restart_csv.xml
PaulTalbot-INL: In order of use, it is an output of this IOStep (to load data into) and as input of the MultiRun (to read data from).  The problem occurs when an object is the output of multiple MultiRun steps.

PaulTalbot-INL:
There is no "fix" currently planned for the Raven Output postprocessor, as its needs have been replaced by the ability for postprocessors to write directly into data objects.  However, there is potential for it to be useful when we find a use case again, for reading general metadata off data objects.  As such, we leave the components here, but remove it from use in RAVEN for now.

PaulTalbot-INL:
I don't have an easy answer for the slower operation of the tests, but I suspect the bulk of it comes from the increased number of running tests.  The new passing tests generally were failing early instead of running to completion; this isn't true any more.

In addition, the fixes for the ExternalPostProcessor in the event the user provides an ND array are notably slower, so that may account for some slowdown.

tests/framework/Samplers/Restart/test_restart_csv.xml
wangcj05: got it.

framework/Samplers/Sampler.py
wangcj05: not sure here, but it is ok to me.

wangcj05:
I think this PR is good.

PR #511, opened 2018-01-12
Asynchronous History Sampling

PaulTalbot-INL:
Fixed adding histories of different lengths through addRealization.

The problem was insufficiently-protected algorithms for determining when an index is identical from one realization to the next.

I added a check for difference in length first, and added this to the unit testing.


PaulTalbot-INL:
It's not currently practical for me to run the tests on falcon before Monday; @wangcj05 if you want to include this change in your working branch, or we can wait for Monday for test change results.

framework/DataObjects/XDataSet.py
```diff
@@ -689,11 +689,16 @@ def _checkAlignedIndexes(self,rlz,tol=1e-15):
     for index in self.indexes:
       # if it's aligned so far, check if it still is
       if index in self._alignedIndexes.keys():
-        # if close enough, then keep the aligned values; otherwise, take action
-        if isinstance(rlz[index][0],(float,int)):
-          closeEnough = all(np.isclose(rlz[index],self._alignedIndexes[index],rtol=tol))
+        # first, if lengths don't match, they're not aligned.
+        if len(rlz[index]) != len(self._alignedIndexes[index]):
+          closeEnough = False
```
wangcj05: I have a question here. It seems we will not check the duplicate data in the indexes if there length are not the same. If I understand correctly, this will increase the memory and slow down the calculation. Let's talk on Monday.

wangcj05: @alfoa I need you comments here.

PaulTalbot-INL: This is correct; if the length of the indexes is not the same, then the old index and the new index _cannot_ be exactly the same, so we have to store them separately.  Right?

wangcj05: As we discussed in the meeting, can you add "FIXME" here regarding the memory and speed issue?

PaulTalbot-INL: done

wangcj05:
This PR is good to me.

framework/DataObjects/XDataSet.py
```diff
@@ -689,11 +689,17 @@ def _checkAlignedIndexes(self,rlz,tol=1e-15):
...
-        if isinstance(rlz[index][0],(float,int)):
-          closeEnough = all(np.isclose(rlz[index],self._alignedIndexes[index],rtol=tol))
+        # first, if lengths don't match, they're not aligned.
+        # TODO there are concerns this check may slow down runs; it should be profiled along with other bottlenecks to optimize our efforts.
+        if len(rlz[index]) != len(self._alignedIndexes[index]):
+          closeEnough = False
         else:
-          closeEnough = all(rlz[index] == self._alignedIndexes[index])
+          # "close enough" if float/int, otherwise require exactness
+          if isinstance(rlz[index][0],(float,int)):
```
alfoa: what about it is a numpy float or int? 
You should add ```numpy.integer``` and ```numpy.number``` for checking all of them

wangcj05: I do have an issue when the ```dtype=np.int32``` in the past.

PaulTalbot-INL:
```python
a = np.array([6],dtype=np.integer)
type(a[0])           # result is <type 'numpy.int64'>
isinstance(a[0],int) # result is True

a = np.array([3.14],dtype=np.float)
type(a[0])             # result is <type 'numpy.float64'>
isinstance(a[0],float) # result is True
```
Inheritance! 😃

wangcj05: Can you test ```np.int32, np.float32, np.int8, ...```?

PaulTalbot-INL: Sorry, why are we using a variety of integer types?  Is this because of different numpy installations?

wangcj05: I have no idea about this, but I do have a case (data mining) when the return type is ```np.int32```, and the code crashes. If I remembered correctly, the type of returned labels is ```np.int32```.

alfoa:
because when the values come from scikitlearn/scipy, they are casted as numpy types

framework/DataObjects/XDataSet.py
alfoa: It is not ```Inheritance```
```np.int64``` and ```np.float64``` are implicitly mapped as python Floats and Integer. I am not surprised your ```isinstance``` works fine with the default ```np.integer``` and ```np.float```.

Indeed : 
```
>>> import numpy as np
>>> a = np.array([6],dtype=np.float32)
>>> type(a[0])
<type 'numpy.float32'>
>>> isinstance(a[0],float)
```

Comment (PaulTalbot-INL) [reply framework/DataObjects/XDataSet.py]: Again, the point is, if we're going to have some dozen types that have no common root but all need to be considered floats, we need a more intelligent way throughout the data objects (and raven in general) to make this kind of a check.  It better be a fast method, too, because looping over the options will be painful.

Is checking against float,int,np.integer,np.float sufficient?  No, because as I just tested, while all the np.int* are np.integer types, not all np.float* are np.float types.  This bears more discussion than comments on this PR merit.

Comment (alfoa) [reply framework/DataObjects/XDataSet.py]: Paul.
Firstly I told you that the correct way to check for any float-like types in numpy is ```numpy.number``` (from which all the numpy.float* inherits from) and not ```numpy.float```
Second, I agree that this checks should be handled in the utilities where we are going to have control on the additional ```types``` we would (if any) need in the future.


```
>>> a = np.array([6],dtype=np.float32)
>>> b = np.array([6],dtype=np.float64)
>>> c = np.array([6],dtype=np.float128)

>>> isinstance(a[0],np.number)
True
>>> isinstance(b[0],np.number)
True
>>> isinstance(c[0],np.number)
True
```

Comment (wangcj05) [reply framework/DataObjects/XDataSet.py]: From NumPy: 
screen shot 2018-01-16 at 12 11 05 pm

Comment (wangcj05) [reply framework/DataObjects/XDataSet.py]: These are the fundamental data types from NumPy

Comment (PaulTalbot-INL) [reply framework/DataObjects/XDataSet.py]: Please, do not add more here.  This needs to be a seperate discussion from this PR.

**Comment by PaulTalbot-INL** [comment]
Test diff results:
```python
**** THE FOLLOWING TESTS CHANGED OUTCOMES ****
**** Test framework.testExternalModel | has changed outcome from <FAILED> to <OK> !!!!
**** Test framework/CodeInterfaceTests.OpenModelicaInterfaceTestNoExecutable | has changed outcome from <FAILED> to <OK> !!!!
**** Test framework/CodeInterfaceTests.testMCMammothNoExecutableBisonRelap7WithHDF5restart | has changed outcome from <FAILED> to <OK> !!!!
*****************
Number of success first file:
398
401
NUMBER OF SUCCESSES INCREASED: +3
TOTAL TIME OF TESTS INCREASED       : reference_time(457.2) < new_time(460.0)
AVERAGE TIMES FOR SUCCESS TESTS ARE : reference_time(1.14874371859) v.s. new_time(1.14713216958)
```

framework/DataObjects/XDataSet.py
PaulTalbot-INL: This is handled in #524.

wangcj05:
The tests should be rerun, since there are some false positive detected in the regression test system.

PaulTalbot-INL:
After merging the test harness fixes into this branch, here is the summary (details in the dev email chain):
```
**** SUMMARY ****
**** THE FOLLOWING TESTS CHANGED OUTCOMES ****
**** Test framework/CodeInterfaceTests.OpenModelicaInterfaceTestNoExecutable | has changed outcome from <FAILED> to <OK> !!!!
**** Test framework/CodeInterfaceTests.testMCMammothNoExecutableBisonRelap7WithHDF5restart | has changed outcome from <FAILED> to <OK> !!!!
*****************
Number of success first file:
383
385
NUMBER OF SUCCESSES INCREASED: +2
```

wangcj05:
Good to go.

PR #526, opened 2018-01-17
rework: test harness fix

PaulTalbot-INL:
Fixes test harness to remove output files before running tests.

Also fixes MC restart case.


PaulTalbot-INL:
It appears we now have a library problem on the linux machines; they can't find pandas.  This should not be possible due to our setup script, but apparently it's happening.

wangcj05:
@PaulTalbot-INL It seems the library issue have existed for a long time (at the beginning of December, 2017).

wangcj05:
The failure is not because of this PR, and I will merge it.

wangcj05:
Let's fix the library issue in a different PR.

PR #538, opened 2018-01-23
rework: CSV fixes and improvements

PaulTalbot-INL:
- Fixes load-multiple-csv tests
 - If a CSV is the output of a multirun step, it will now append instead of rewrite each time


PaulTalbot-INL:
That's a good question.  It will definitely crash, but I don't know that it will crash with an intelligent error.  I can make a fail-test for it, probably.

framework/DataObjects/XDataSet.py
```diff
@@ -1124,7 +1124,7 @@ def _fromCSV(self,fileName,**kwargs):
         data = df[[var,self.sampleTag]].groupby(self.sampleTag).first().values[:,0]
         dtype = self._getCompatibleType(data.item(0))
         arrays[var] = self._collapseNDtoDataArray(data,var,labels=samples,dtype=dtype)
-    self._convertArrayListToDataset(arrays,action='replace')
+    self._convertArrayListToDataset(arrays,action='extend')
```
alfoa: where do you check if the new data are consistent with the old ones? It might cause an unexpected behavior

alfoa: Oh...I see Congjian and you already found the issue

PaulTalbot-INL: The error isn't too bad, actually, it says that the expected variables weren't found found in <filename>.

PaulTalbot-INL:
Added test passes, no code changed:
```
framework/DataObjects.test_output_attributes.... skipped (skipped (Missing executable: "../../../RAVEN-opt"))
framework/DataObjects.load_csv_history.................................................................... OK
framework/DataObjects.load_two_csvs....................................................................... OK
framework/DataObjects.load_two_csv_histories_fail......................................................... OK
framework/DataObjects.load_two_csv_histories.............................................................. OK
framework/DataObjects.dataObject_metadata................................................................. OK
framework/DataObjects.load_csv............................................................................ OK
-------------------------------------------------------------------------------------------------------------
Ran 6 tests in 7.0 seconds
6 passed, 1 skipped, 0 pending, 0 failed
```

wangcj05:
Can you add a unit test for this?

wangcj05:
In addition, could you report the unit tests results for the DataObject?

PaulTalbot-INL:
I think this would be a poor application for the unit test; it's testing a RAVEN workflow, which is better tested by the framework test.

The unit tests are all in the regression tests system, and they're not in the failing test list, ergo they're passing in both the reference and new case.  What would you like to see from running the unit tests independently?

wangcj05:
First, you add an extend capability, this should be simple to added to the unit test, and would be simple for the future maintenance. Second, you modified the DataObject, but the tests reported here are only from the raven framework tests, not the unit tests. In addition, if you decide to add addition unit test to test the extend capability, the test results should be also reported here.

wangcj05:
Talked directly with @PaulTalbot-INL, the mechanics is already tested in the unit test. And this PR looks good to me.

wangcj05:
Any comments @alfoa ?

PR #548, opened 2018-01-25
rework: Raven-runs-raven with ROM

PaulTalbot-INL:
Fixes the original raven-runs-raven test.

Also fixes a potential bug that ignores data when _one of_ the targets of an SKL ROM is consistently the same, but the others are not.  Adds a test to check the same.


PaulTalbot-INL:
Tests changed:
```
**** SUMMARY ****
**** Unmatched tests are:

framework/ROM.SKL_time
**** THE FOLLOWING TESTS CHANGED OUTCOMES ****
**** Test framework/CodeInterfaceTests.RAVENrunningRAVEN_ROM | has changed outcome from <FAILED> to <OK> !!!!
*****************
**** THE FOLLOWING TESTS HAVE BEEN ADDED IN 2ND FILE ****
**** Test framework/ROM.SKL_time | has been ADDED  ! It is present in 2nd file BUT NOT in 1st file !!!!
*****************
Number of success first file:
466
467
NUMBER OF SUCCESSES INCREASED: +1
**** NUMBER OF TESTS IN THE TWO FILES ARE DIFFERENT: ****
**** first file :536 /= second file: 537****
```
For the record, new test SKL_time also passes in this branch; that's not totally clear from the summary.

framework/LearningGate.py
```diff
@@ -150,10 +150,9 @@ def train(self,trainingSet):
         # the check on the number of time steps (consistency) is performed inside the historySnapShoots method
         # get the time slices
         newTrainingSet = mathUtils.historySnapShoots(trainingSet, len(self.historySteps))
-        if type(newTrainingSet).__name__ != 'list':
-          self.raiseAnError(IOError,newTrainingSet)
+        assert(type(newTrainingSet).__name__ == 'list')
         # copy the original ROM
-        originalROM = copy.deepcopy(self.supervisedContainer[0])
+        originalROM = self.supervisedContainer[0]
```
wangcj05: Could you tell me the reason why the deep copy is removed?

tests/framework/CodeInterfaceTests/gold/raven_running_raven_internal_models/testPointSet_dump.csv
```diff
@@ -1,16 +1,16 @@
...
+4.14951117603,2000.0,9.63150662775e-05,128.124447802,507223.692576
+3.9482296471,0.0,0.000103102494869,146.755892585,507034.274649
+3.9482296471,1000.0,0.000103102494869,146.755892585,507129.874665
+3.9482296471,2000.0,0.000103102494869,146.755892585,507225.47468
+3.92810601029,0.0,9.86735092321e-05,396.580726233,507058.17048
+3.92810601029,1000.0,9.86735092321e-05,396.580726233,507153.770495
+3.92810601029,2000.0,9.86735092321e-05,396.580726233,507249.370511
+3.91082378303,0.0,9.97152314658e-05,345.56072516,507053.2904
+3.91082378303,1000.0,9.97152314658e-05,345.56072516,507148.890415
+3.91082378303,2000.0,9.97152314658e-05,345.56072516,507244.49043
```
wangcj05: Could you tell me the reason for regolding?

tests/framework/ROM/TimeDepSKL/projectile.py
```diff
@@ -0,0 +1,129 @@
...
+    self.r[i] = r
+  self.time = ts
+
+#can be used as a code as well
+if __name__=="__main__":
+  import sys
+  inFile = sys.argv[sys.argv.index('-i')+1]
+  outFile = sys.argv[sys.argv.index('-o')+1]
+  #construct the input
+  Input = {}
```
wangcj05: I think we need to change the name here.

tests/framework/CodeInterfaceTests/gold/raven_running_raven_internal_models/testPointSet_dump.csv
PaulTalbot-INL: The change is in the 5th and higher digit in most cases, and the model is quite sensitive to small changes in precision, so I feel it is close enough as to be the same, given that this is not an analytic test.

tests/framework/ROM/TimeDepSKL/projectile.py
PaulTalbot-INL: I disagree, I think developers have autonomy within their external models.

framework/LearningGate.py
PaulTalbot-INL: If you look below, each time the originalROM is used, it's surrounded in a deepcopy; therefore, we're wasting a deepcopy (which is expensive) by deepcopying here.

wangcj05: ok

wangcj05:
I will merge this, just keep in mind that we have some issue with test, the accuracy of outputs is changed without a good reason.

PR #570, opened 2018-02-05
CSV printing speedup

PaulTalbot-INL:
After some testing, we no longer can find any problem with the original way of writing to CSV; this is possibly because the user cannot get "RAVEN_sample_ID" in CSV outputs anymore.

While the "garbled" indexes might return in the future, for now this speeds up the writing of DataSets generally.

Also I apparently removed a bunch of trailing whitespace from some doc files.


PaulTalbot-INL:
It looks like just the melcor interface is failing on Linux.  Is that an outstanding problem or an expected fail currently?

alfoa:
```framework.testExternalModel FAILED (CWD /opt/civet/build_0/raven/tests/framework/Samplers/AdaptiveBatch METHOD opt Expected files not created /opt/civet/build_0/raven/tests/framework/externalModel/testPrintHistorySet_dump_0.csv /opt/civet/build_0/raven/tests/framework/externalModel/testPrintHistorySet_dump_1.csv /opt/civet```
O_O

PaulTalbot-INL:
I'm not understanding why this is behaving different on subsequent tests.

PR #586, opened 2018-02-27
Job Profiling

PaulTalbot-INL:
This PR adds individual job profiling to RAVEN's intrinsic abilities.  Since rough print-based timing schemes have been implemented by myself and several other developers separately at different times, it seems this is a common need for profiling job distribution.

In the future, this same profiling approach can be extended to BaseClass raven entities, and other timing statements can be added to jobs without significant effort.  Also, when a log file is implemented these timings could be useful there.

To use this feature, add `profile="jobs"` to the `Simulation` node, and on each job's deconstruction, it will print a summary of its timings.


framework/JobHandler.py
```diff
@@ -349,6 +355,7 @@ def reAddJob(self, runner):
         self.__queue.append(runner)
       else:
         self.__clientQueue.append(runner)
+      runner.trackTime('queue')
```
alfoa: shouldn't we track the time only in case the profile flag is True?

framework/JobHandler.py
```diff
@@ -672,8 +681,17 @@ def cleanJobQueue(self):
           ## it by calling numRunning.
           with self.__queueLock:
             self.__finished.append(run)
+            self.__finished[-1].trackTime('jobHandler_finished')
             runList[i] = None
 
+  def setProfileJobs(self,profile=True):
```
alfoa: shouldn't be ```False``` by default?

framework/JobHandler.py
PaulTalbot-INL: I can probably make that happen, yeah.

framework/JobHandler.py
PaulTalbot-INL: My thinking was that if the method is being called, then you mean to set it to "True"; I can easily reverse this though.

PaulTalbot-INL:
This needs to rerun tests after #595 is merged.

PR #599, opened 2018-03-21
cleaned up contribution

alfoa:
Just cleaned up commented code in PP.

### DIFF
--- a/framework/contrib/pp/pp.py
+++ b/framework/contrib/pp/pp.py
@@ -130,7 +130,7 @@ class _Worker(object):
     """
     command = [sys.executable, "-u", "-m", "ppworker"]
 
-    command.append("2>/dev/null")
+    #command.append("2>/dev/null")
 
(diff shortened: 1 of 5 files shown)


framework/contrib/pp/pp.py
```diff
@@ -541,37 +536,33 @@ def get_stats(self):
...
-        return returnList
-
-
-    def print_stats(self):
-        """Prints job execution statistics. Useful for benchmarking on
-           clusters"""
-        for row in self.collect_stats_in_list(): print(row)
+            print "WARNING: statistics provided above is not accurate" \
+                  " due to job rescheduling"
+        print
```
PaulTalbot-INL: Are the new prints in this file meant to reside in the code, or are they the result of debugging work?

derekstucki: Print without parens is a python 3 incompatibility. Parens also eliminate the need to escape line breaks.

alfoa:
Not needed.

PR #600, opened 2018-03-21
Optimizer inherits from Sampler

PaulTalbot-INL:
Makes most of the Sampler base methods inherited in the Optimizer instead of re-implementing.


PaulTalbot-INL:
Expected tests pass. Merging.

PR #601, opened 2018-03-27
changed the printing strategy of profiles

alfoa:
The ```profile``` info in the job handler are printed by the Runners (i.e. threads/cpus). This can cause that in case of simultaneous printing, not all the information of a job are printed sequentially causing the impossibility to connect the info reported with the actual job. This PR improves this behavior creating a string a-priori and print it all in once.

### DIFF
--- a/framework/JobHandler.py
+++ b/framework/JobHandler.py
@@ -33,7 +33,6 @@ import os
 import copy
 import sys
 import abc
-#import logging, logging.handlers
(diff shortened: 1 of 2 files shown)


alfoa:
I created a script. But there are a lot of workaround for the not-thread safe printing. I will modify the script and commit it (in another PR) once this is in.

PR #615, opened 2018-04-17
Improved UnorderedCSVDiffer speed

PaulTalbot-INL:
Marginally increases the speed of the UnorderedCSVDiffer (5-10% faster).

Also adds external model test as prereq for external model reseed test, which checks the same files and sometimes caused a missing CSV error.


tests/framework/tests
```diff
@@ -209,6 +209,7 @@
   type = 'RavenFramework'
   input = 'test_external_reseed.xml'
   output = 'externalModel/testPrintHistorySet_dump_0.csv externalModel/testPrintHistorySet_dump_1.csv externalModel/testPrintHistorySet_dump_2.csv'
+  prereq = testExternalModel
```
alfoa: This modification is already in dataobject-rework ...why no conflict is shown??? O_o

PaulTalbot-INL: If the change is identical, I think maybe it can handle coinciding both effects.

PaulTalbot-INL: Let me rebase to assuage those concerns.

PaulTalbot-INL:
That didn't work as desired.   A moment...

PR #618, opened 2018-04-19
expand install script for conda 4.4 and beyond

PaulTalbot-INL:
Extends the "establish_conda_env.sh" script to work for conda versions after 4.3, where making conda available is a function definition process instead of a path modification.

This is necessary because:
 1. `which` doesn't find `conda` since it isn't on `$PATH`
 1. `conda` itself can't be used to install libraries without sourcing `conda.sh`, since functions are not transferred from environment to shell script.

Closes #618

### DIFF
--- a/scripts/establish_conda_env.sh
+++ b/scripts/establish_conda_env.sh
@@ -30,6 +30,9 @@ try_using_raven_environment ()
     fi
 }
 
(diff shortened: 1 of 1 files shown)


PaulTalbot-INL:
Can someone review this so we can consider merging it?

PR #632, opened 2018-05-01
Vector constants

PaulTalbot-INL:
Gives constants the power to be vectors, too.

Constants can have the "shape" parameter in order to shape comma-separated values as per `numpy.reshape`, or if omitted will be a vector of length equal to the number of entires, unless that length is 1, then it will be single-valued float as before.

To accomodate this, "shape" is moved from a node to  an attribute for `variables`, and for clarity we generalized the reading of constants to be inherited from `Sampler` to `Optimizer`.

We also add two new `InputData` parameter types: the `FloatList` and `IntegerList`, which expect pretty much exactly what their names say.


PaulTalbot-INL:
I have one more merge to add to this, to extend the raven-running-raven interface to handle the vectors.

framework/CodeInterfaces/RAVEN/RAVENInterface.py
```diff
@@ -80,26 +80,40 @@ def _readMoreXML(self,xmlNode):
...
+        if checkImport is None:
+          raise IOError(self.printTag+' ERROR: the conversionModule "{}" failed on import!'
+                        .format(self.extModForVarsManipulationPath))
+        # check methods are in place
+        noScalar = 'convertNotScalarSampledVariables' in checkImport.__dict__
+        scalar = 'manipulateScalarSampledVariables' in checkImport.__dict__
+        if not (noScalar or scalar):
+          raise IOError(self.printTag +' ERROR: the conversionModule "'+source
+                                      +'" does not contain any of the usable methods! Expected at least '
+                                      +'one of: "manipulateScalarSampledVariables" and/or "manipulateScalarSampledVariables"!')
```
alfoa: this line is 128 characters....the maximum should be 120 :)

alfoa: (I know it is from previous code :D )

framework/Samplers/Sampler.py
```diff
@@ -55,6 +55,7 @@ class cls.
         specifying input of cls.
     """
     inputSpecification = super(Sampler, cls).getInputSpecification()
+    # FIXME the DET HybridSampler doesn't use the "name" param for the samples it creates, so we can't require the name yet
```
alfoa: line is 124 columns... we should split this.

PS. Can you remember to give me a bit more details  in 30 mins about this thing?

Issue #84, opened 2017-04-06
Improve RAVEN input handling 


joshua-cogliati-inl:
This task is to improve RAVEN's input handling.

Goals:

Provide enough information in the class to be able to:

 *   Generate a GUI (like Peacock) from the data.
 *   Check that the input is valid
 *   Generate an xsd file.

The information should be generated statically (from static methods). The suggested method is to create functions in each class that have the correct data, and use that data to generate the needed information.

This is being done through the utils.InputData system which needs to be extended to more classes.

Originally: https://hpcgitlab.inl.gov/idaholab/raven/issues/611

We need to re-standardize the objects that are using the Input Data following the instructions reported in https://github.com/idaholab/raven/wiki/input-data
. An example is the Sampler, where the InputData is used within the read more XML

joshua-cogliati-inl:
After Input improvements #395 goes in, PostProcessors, Steps, Distributions and DataObjects use the input handling system.  

Samplers use the input handling system to parse the input, but generally still use reading the xml to find the data inside.

If you want to check if a specific class use it, check to see if there is a readMoreXML function.  If there is, then it should use something like:
```
  def _readMoreXML(self,xmlNode):
    """
      Function to read the xml input block.
      @ In, xmlNode, xml.etree.ElementTree.Element, xml node
      @ Out, None
    """
    paramInput = self.getInputSpecification()()
    paramInput.parseNode(xmlNode)
    self._handleInput(paramInput)
```
if the input handling system is being used.

There are several subsystems not yet using the input handling system.   These include the Databases, Models, Files and RunInfo.  The Samplers need to use the input handling system for reading in data and not just parse the data.  Note that RunInfo and Models allow plugins to handle raw xml, so either the interface needs to be changed for the plugins, or a way to keep track of the original xml is needed.

wangcj05:
@idaholab/raven-team Hello All, I think this is the issue for the InputData, let's push to further employment of InputData of our development. Please keep update this issue during the future development.

PaulTalbot-INL:
I don't see a list of where it's not in place yet; does that information exist? It will be challenging to carve out other work to include that without knowing what needs to be targeted.

wangcj05:
> I don't see an updated list of where it's not in place yet; does that information exist? It will be challenging to carve out other work to include that without knowing what needs to be targeted.

I will check our code, and list the things that we need to convert to use InputData.

PR #1332, opened 2020-10-08
Add capability to handle unknown subs at runtime

dylanjm:
--------
Pull Request Description
--------

Refs idaholab/HERON#38 and #1335 

By adding an `unknownSubs` boolean value to the utils::InputParameter class we
can handle InputParameter classes that may have an unknown name or number of
variables.

This is handled by checking the `unknownSubs` boolean during parsing. If `True`
it will add the sub to the class on the fly.

This only applies to subparts on input parameters and will default to using
`InputTypes.InputString` as its' type.

See the referenced issue for a use-case. This could be helpful for users who
want to feed-forward static data through the RAVEN stack.

----------------
----------------

### DIFF
--- a/framework/utils/InputData.py
+++ b/framework/utils/InputData.py
@@ -122,6 +122,7 @@ class ParameterInput(object):
   name = "unknown"
   subs = OrderedDict() #set()
   subOrder = None
+  unknownSubs = False  # Allow unknown subs at run-time
   parameters = OrderedDict()
   contentType = None
   strictMode = True #If true, only allow parameters and subnodes that are listed
@@ -143,7 +144,7 @@ class ParameterInput(object):
 
   @classmethod
   def createClass(cls, name, ordered=False, contentType=None, baseNode=None,
-                  strictMode=True, descr=None, printPriority=None):
+                  strictMode=True, descr=None, printPriority=None, unknownSubs=False):
     """
       Initializes a new class.
       @ In, name, string, The name of the node.
@@ -184,6 +185,7 @@ class ParameterInput(object):
       cls.parameters = {}
       cls.subs = OrderedDict() #set()
       cls._subDict = dict()
+      cls.unknownSubs = unknownSubs
       if ordered:
         cls.subOrder = []
       else:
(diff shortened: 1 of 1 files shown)


PaulTalbot-INL:
How does this interact with the automatic documentation generation?

Also, there may be overlap between `strict=False` and `unknownSubs=True`. @joshua-cogliati-inl @alfoa if we have strict mode on False for an InputData node, that implies accepting anything under it but doesn't actually read them into the specs, right? Could we use Dylan's fix based on the strict mode instead of with another flag, or does strict mode do something else too?

dylanjm:
> How does this interact with the automatic documentation generation? 

@PaulTalbot-INL I was thinking the same thing. Since these nodes would be unknown there would be no way to officially document them. What's nice is that you can still add documented subnodes to the class, it should be mentioned in the documentation that unknown subnodes will be treated as constants always... or something like that.

alfoa:
An issue in the RAVEN repository needs to be created here.

PaulTalbot-INL:
> An issue in the RAVEN repository needs to be created here.

Can an issue in a plugin not be justification for a PR in RAVEN? Not that we don't have this problem in RAVEN (see #84, #936), but in principle, does the issue have to be in RAVEN itself, philosophically/technically speaking?

alfoa:
If I understood correctly, this modification would allow for misspelling nodes (that are supposed to be checked). Is it correct? 

For example, If I have a node supposed to call "myNode" and I call it "myNose", the node will be accepted as "unknown" and fail later on?

Is it right?

alfoa:
> > An issue in the RAVEN repository needs to be created here.
> 
> Can an issue in a plugin not be justification for a PR in RAVEN? Not that we don't have this problem in RAVEN (see #84, #936), but in principle, does the issue have to be in RAVEN itself, philosophically/technically speaking?

I think we need to have it in RAVEN too (referencing to the plugin issue). Also because we require to test this modification (aka unit testing in this case)

dylanjm:
> For example, If I have a node supposed to be called "myNode" and I call it "myNose", the node will be accepted as "unknown" and fail later on?

@alfoa That should be the default behavior that is caught at run-time and an error showing the user `myNose` is not an allowed parameter. 

This modification would work when users want to add custom subnodes that we, as developers, might not know before hand. In the referenced HERON issue the example is:

We, the developers, add an optional `case_info` node that can be populated with static information to help identify runs with associated output data. Such as:

```xml
<Case>
...
<case_info>
  <state>California</state>
  <mylastname>McDowell</mylastname>
...
</case_info>
```

We can't declare `state` and `mylastname` as InputParameters so we provide a way to let a Node have an unknown number of subnodes just as long as they don't mean anything to actual model computation.

dylanjm:
After talking with @PaulTalbot-INL we are going to go a different direction with handling this static information. It does appear `strict` and `unknownSubs` are duplicates and can probably be phased out eventually. I'll go ahead and close this PR.

PR #1422, opened 2021-01-29
Update issue templates

alfoa:
Added some names for the proposed assignees

### DIFF
--- a/.github/ISSUE_TEMPLATE/defect-report.md
+++ b/.github/ISSUE_TEMPLATE/defect-report.md
@@ -3,7 +3,7 @@ name: Defect Report
 about: Report a DEFECT you experienced using the code
 title: "[DEFECT] Title Here"
 labels: defect, priority_normal
-assignees: ''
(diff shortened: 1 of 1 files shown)


wangcj05 (approved):
No code change, just update issue template.

wangcj05:
@alfoa where is the related issue, the PR description and the checklist?

PaulTalbot-INL:
I don't love auto-assignment; I like to be assigned to issues that I will be handling immediately, not a backlog of assignments that I may or may not ever get to. I think no default assignees is much better than having everyone assigned to every issue; we already have problems with multiple people fixing the same issue at once as it is.

alfoa:
Unfortunately i released that "no default" determined that there are several issues that we did not even "reviewed". The initial default assignment is to notify (Forcefully) everyone that a new Defect is out there (I changed this for defects only...I hope we do not get 10 issues per day). what do you think?
Andrea

Issue #1485, opened 2021-03-17
[DEVELOPMENT] TimeSeriesAnalysis algorithms


PaulTalbot-INL:
--------
Under Discussion Topic
--------

**Background of the development decision**
Frequently in RAVEN entities (`ROM`, `Sampler`, `PostProcessor`, etc) we need to use algorithms embedded in other entities. Frequently for entities needing to use ROMs, this requires instantiating, initializing, using, and deconstructing the embedded entity, forcing the using entity to implement the API that is typically reserved for the `Step`.

In particular, the ARMA, Fourier, PeakPicking, etc. embedded in the `SupervisedLearning.ARMA` object can be useful in characterizing time series in other entities such as PostProcessors.

In successive development meetings in the end of 2020 and early 2021, the decision was made to create a `TimeSeriesAnalysis` (TSA) module, containing algorithms for time series analysis that are not specific to any given RAVEN entity, such as ARMA, Fourier, etc. The `SupervisedLearning.ARMA` class can then be rewritten as a `SupervisedLearning.SyntheticHistory` class that leverages the TSA module algorithms. This allows other entities such as time series characterization PostProcessors to also use the `TSA` algorithms without instantiating `SupervisedLearning.SyntheticHistory` objects.

**Broader development implication**
This decision sets precedent for preferring algorithms be separated from entities when the algorithms are needed in multiple parts in the code.

If the algorithms are sufficiently straightforward, they might be included in the `utils` libraries, particularly `mathUtils`.

If the algorithms are a family of related algorithms, it may be preferable to create a new module for that family of algorithms.

**References**
PRs for TSA:
- Implementations: #1434, #1437, #1440, #1444  
- Introduction: #1405

----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [x] 1. Is it tagged with the under_discussion type?
- [x] 2. If implemented, it will add a new requirement?
- [x] 3. Is a rationale provided? (Such as explaining why the improvement is needed )

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [x] 1. The discussion determined the addition of a new ```task``` issue?

dylanjm:
@PaulTalbot-INL Should I close #1443 or would you consider this a different "issue". One addressing the [Task] of adding algos and one addressing the [Development Decisions]?

PaulTalbot-INL:
[...]

Issue #1404, opened 2021-01-12
[TASK] Modularization of ARMA ROM


PaulTalbot-INL:
--------
Issue Description
--------

The ARMA ROM has grown to the point where it could be modularized to great benefit.

Current Time Series Analysis (TSA) algorithms embedded in the ARMA ROM include:
- Fourier detrending
- ARMA characterization (including VARMA)
- Peak characterization

Further modularity could be introduced to allow rearranging, addition, and selection of these algorithms, both improving the usability, expanding the feature set, and improving the maintainability of the algorithms for synthetic history generation.

This is needed in particular for the signal analysis applications in the FY21-23 cyber LDRD.

----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [x] 1. Is it tagged with a type: defect or task?
- [x] 2. Is it tagged with a priority: critical, normal or minor?
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements?
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [ ] 1. If the issue is a defect, is the defect fixed?
- [ ] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [ ] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [ ] 4. If the issue is a defect, does it impact the latest release branch? If yes, is there any issue tagged with release (create if needed)?
- [ ] 5. If the issue is being closed without a pull request, has an explanation of why it is being closed been provided?

PR #1488, opened 2021-03-23
Add modular inheritence structure for TSA module

dylanjm:
--------
Pull Request Description
--------

Refs #1443 #1485 #1404

* Add TimeSeriesGenerator class
* Add TimeSeriesCharacterizer class
* Add predicate functions to base TSA class
* Modify unit-tests

----------------
----------------

### DIFF
--- a/framework/SupervisedLearning/SyntheticHistory.py
+++ b/framework/SupervisedLearning/SyntheticHistory.py
@@ -134,6 +134,8 @@ class SyntheticHistory(supervisedLearning):
       targets = settings['target']
       indices = tuple(self.target.index(t) for t in targets)
       params = self.trainedParams[algo]
+      if not algo.canGenerate():
+        self.raiseAnError(IOError, "This TSA algorithm cannot generate synthetic histories.")
       signal = algo.generate(params, pivots, settings)
       result[:, indices] += signal
     # RAVEN realization construction
--- a/framework/TSA/ARMA.py
+++ b/framework/TSA/ARMA.py
@@ -25,11 +25,11 @@ from utils import InputData, InputTypes, randomUtils, xmlUtils, mathUtils, impor
 statsmodels = importerUtils.importModuleLazy('statsmodels', globals())
 
 import Distributions
-from .TimeSeriesAnalyzer import TimeSeriesAnalyzer
+from .TimeSeriesAnalyzer import TimeSeriesAnalyzer, TimeSeriesGenerator, TimeSeriesCharacterizer
 
 
 # utility methods
-class ARMA(TimeSeriesAnalyzer):
+class ARMA(TimeSeriesGenerator, TimeSeriesCharacterizer):
   r"""
     AutoRegressive Moving Average time series analyzer algorithm
   """
@@ -76,7 +76,7 @@ class ARMA(TimeSeriesAnalyzer):
       @ Out, None
     """
     # general infrastructure
-    TimeSeriesAnalyzer.__init__(self, *args, **kwargs)
+    super().__init__(self, *args, **kwargs)
     self._minBins = 20 # this feels arbitrary; used for empirical distr. of data
 
   def handleInput(self, spec):
--- a/framework/TSA/ARMA.py
+++ b/framework/TSA/ARMA.py
@@ -25,7 +25,7 @@ from utils import InputData, InputTypes, randomUtils, xmlUtils, mathUtils, impor
 statsmodels = importerUtils.importModuleLazy('statsmodels', globals())
 
 import Distributions
-from .TimeSeriesAnalyzer import TimeSeriesAnalyzer, TimeSeriesGenerator, TimeSeriesCharacterizer
+from .TimeSeriesAnalyzer import TimeSeriesGenerator, TimeSeriesCharacterizer
 
 
 # utility methods
@@ -85,7 +85,7 @@ class ARMA(TimeSeriesGenerator, TimeSeriesCharacterizer):
       @ In, inp, InputData.InputParams, input specifications
       @ Out, settings, dict, initialization settings for this algorithm
     """
-    settings = TimeSeriesAnalyzer.handleInput(self, spec)
+    settings = super().handleInput(spec)
     settings['P'] = spec.findFirst('SignalLag').value
     settings['Q'] = spec.findFirst('NoiseLag').value
     settings['reduce_memory'] = spec.parameterValues.get('reduce_memory', settings['reduce_memory'])
--- a/framework/TSA/ARMA.py
+++ b/framework/TSA/ARMA.py
@@ -98,7 +98,7 @@ class ARMA(TimeSeriesGenerator, TimeSeriesCharacterizer):
       @ In, settings, dict, existing settings
       @ Out, settings, dict, modified settings
     """
-    settings = TimeSeriesAnalyzer.setDefaults(self, settings)
+    settings = super().setDefaults(settings)
     if 'gaussianize' not in settings:
       settings['gaussianize'] = True
     if 'engine' not in settings:
--- a/framework/TSA/ARMA.py
+++ b/framework/TSA/ARMA.py
@@ -76,7 +76,7 @@ class ARMA(TimeSeriesGenerator, TimeSeriesCharacterizer):
       @ Out, None
     """
     # general infrastructure
-    super().__init__(self, *args, **kwargs)
+    super().__init__(*args, **kwargs)
     self._minBins = 20 # this feels arbitrary; used for empirical distr. of data
 
   def handleInput(self, spec):
--- a/framework/TSA/PolynomialRegression.py
+++ b/framework/TSA/PolynomialRegression.py
@@ -20,10 +20,10 @@ import utils.importerUtils
 statsmodels = utils.importerUtils.importModuleLazy("statsmodels", globals())
 
 from utils import InputData, InputTypes, randomUtils, xmlUtils, mathUtils, utils
-from .TimeSeriesAnalyzer import TimeSeriesAnalyzer
+from .TimeSeriesAnalyzer import TimeSeriesAnalyzer, TimeSeriesCharacterizer, TimeSeriesGenerator
 
 
-class PolynomialRegression(TimeSeriesAnalyzer):
+class PolynomialRegression(TimeSeriesGenerator, TimeSeriesCharacterizer):
   """
   """
 
@@ -53,7 +53,7 @@ class PolynomialRegression(TimeSeriesAnalyzer):
       @ Out, None
     """
     # general infrastructure
-    TimeSeriesAnalyzer.__init__(self, *args, **kwargs)
+    super().__init__(self, *args, **kwargs)
 
   def handleInput(self, spec):
     """
--- a/framework/TSA/PolynomialRegression.py
+++ b/framework/TSA/PolynomialRegression.py
@@ -36,7 +36,7 @@ class PolynomialRegression(TimeSeriesGenerator, TimeSeriesCharacterizer):
         specifying input of cls.
     """
     specs = super(PolynomialRegression, cls).getInputSpecification()
-    specs.name = 'regression'
+    specs.name = 'PolynomialRegression'
     specs.description = """TimeSeriesAnalysis algorithm for fitting data of degree one or greater."""
     specs.addSub(InputData.parameterInputFactory('degree', contentType=InputTypes.IntegerType,
                                                  descr="Specifies the degree polynomial to fit the data with."))
@@ -53,7 +53,7 @@ class PolynomialRegression(TimeSeriesGenerator, TimeSeriesCharacterizer):
       @ Out, None
     """
     # general infrastructure
-    super().__init__(self, *args, **kwargs)
+    super().__init__(*args, **kwargs)
 
   def handleInput(self, spec):
     """
--- a/framework/TSA/PolynomialRegression.py
+++ b/framework/TSA/PolynomialRegression.py
@@ -20,7 +20,7 @@ import utils.importerUtils
 statsmodels = utils.importerUtils.importModuleLazy("statsmodels", globals())
 
 from utils import InputData, InputTypes, randomUtils, xmlUtils, mathUtils, utils
-from .TimeSeriesAnalyzer import TimeSeriesAnalyzer, TimeSeriesCharacterizer, TimeSeriesGenerator
+from .TimeSeriesAnalyzer import TimeSeriesCharacterizer, TimeSeriesGenerator
 
 
 class PolynomialRegression(TimeSeriesGenerator, TimeSeriesCharacterizer):
@@ -61,7 +61,7 @@ class PolynomialRegression(TimeSeriesGenerator, TimeSeriesCharacterizer):
       @ In, inp, InputData.InputParams, input specifications
       @ Out, settings, dict, initialization settings for this algorithm
     """
-    settings = TimeSeriesAnalyzer.handleInput(self, spec)
+    settings = super().handleInput(spec)
     settings['degree'] = spec.findFirst('degree').value
     return settings
 
--- a/framework/TSA/TimeSeriesAnalyzer.py
+++ b/framework/TSA/TimeSeriesAnalyzer.py
@@ -35,6 +35,7 @@ class TimeSeriesAnalyzer(utils.metaclass_insert(abc.ABCMeta, object)):
     """
     A predicate function to determine if object instance inherits from TimeSeriesGenerator.
 
+    @ In, None
     @ Out, boolean, True if instance can generate
     """
     return isinstance(self, TimeSeriesGenerator)
@@ -43,6 +44,7 @@ class TimeSeriesAnalyzer(utils.metaclass_insert(abc.ABCMeta, object)):
     """
     A predicate function to determine if object instance inherits from TimeSeriesCharacterizer.
 
+    @ In, None
     @ Out, boolean, True if instance can characterize
     """
     return isinstance(self, TimeSeriesCharacterizer)
@@ -127,15 +129,6 @@ class TimeSeriesGenerator(TimeSeriesAnalyzer):
   Act as an identifying class for algorithms that can generate synthetic time histories.
   """
 
-  def __init__(self, *args, **kwargs):
-    """
-      A constructor that will appropriately intialize a supervised learning object
-      @ In, args, list, an arbitrary list of positional values
-      @ In, kwargs, dict, an arbitrary dictionary of keywords and values
-      @ Out, None
-    """
-    self.name = self.__class__.__name__ # the name the class shall be known by during its RAVEN life
-
   @abc.abstractmethod
   def generate(self, params, pivot, settings):
     """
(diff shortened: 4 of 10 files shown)


framework/TSA/ARMA.py
```diff
@@ -25,11 +25,11 @@
 statsmodels = importerUtils.importModuleLazy('statsmodels', globals())
 
 import Distributions
-from .TimeSeriesAnalyzer import TimeSeriesAnalyzer
+from .TimeSeriesAnalyzer import TimeSeriesAnalyzer, TimeSeriesGenerator, TimeSeriesCharacterizer
```
PaulTalbot-INL, line 28: We can omit the base import here, right? Similarly for the other TSA files.

framework/TSA/TimeSeriesAnalyzer.py
```diff
@@ -31,13 +31,28 @@ class TimeSeriesAnalyzer(utils.metaclass_insert(abc.ABCMeta, object)):
   ## define the clusterable features for this trainer.
   _features = []
 
+  def canGenerate(self) -> bool:
+    """
+    A predicate function to determine if object instance inherits from TimeSeriesGenerator.
+
+    @ Out, boolean, True if instance can generate
```
PaulTalbot-INL, line 38: technically the RAVEN standards still require a `@ In, None` if there's no input arguments. Applies to several new functions.

framework/TSA/TimeSeriesAnalyzer.py
```diff
@@ -109,6 +112,30 @@ def getResidual(self, initial, params, pivot, settings):
...
+  """
+
+  def __init__(self, *args, **kwargs):
+    """
+      A constructor that will appropriately intialize a supervised learning object
+      @ In, args, list, an arbitrary list of positional values
+      @ In, kwargs, dict, an arbitrary dictionary of keywords and values
+      @ Out, None
+    """
+    self.name = self.__class__.__name__ # the name the class shall be known by during its RAVEN life
```
PaulTalbot-INL, line 137: This appears to be common to both subclasses of TSA; can it remain in the base class?

PaulTalbot-INL (changes requested):
A few comments to address.

framework/TSA/ARMA.py
dylanjm: I believe we need it for the `handleInput()` method we use the base class to call that method.

framework/TSA/TimeSeriesAnalyzer.py
dylanjm: Yes, I believe so.

framework/TSA/ARMA.py
PaulTalbot-INL: Interesting. Would this work as a super call instead of directly calling TimeSeriesAnalyzer.handleInput, or does that cause inheritance resolution issues?

dylanjm:
@PaulTalbot-INL I think this needs a fresh review as well.

framework/SupervisedLearning/SyntheticHistory.py
```diff
@@ -129,7 +129,10 @@ def __evaluateLocal__(self, featureVals):
       targets = settings['target']
       indices = tuple(self.target.index(t) for t in targets)
       params = self.trainedParams[algo]
-      signal = algo.generate(params, pivots, settings)
+      if algo.canGenerate():
+        signal = algo.generate(params, pivots, settings)
+      else:
+        signal = algo.signal;
```
PaulTalbot-INL, line 135: This might need thinking about. I'm not sure how much we want to just return the original signal if it can't generate. Technically if a TSA can't generate, it shouldn't be usable by SyntheticHistory.

I wonder if we need a Postprocessor that acts like the SyntheticHistory but only extracts characteristics and puts them into a DataObject for each realization. This would allow testing characterizers rather than generators.

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+  Auto Correlation Transformation for Characterization of Time-Series Data.
+"""
+import numpy as np
+
+from utils import InputData, InputTypes, xmlUtils
+from .TimeSeriesAnalyzer import TimeSeriesGenerator, TimeSeriesCharacterizer
+
+
+# utility methods
+class AutoCorrelation(TimeSeriesGenerator , TimeSeriesCharacterizer):
```
PaulTalbot-INL, line 24: This guy isn't a generator, right?

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+
+  @classmethod
+  def getInputSpecification(cls):
+    """
+      Method to get a reference to a class that specifies the input data for class cls.
+      @ Out, specs, InputData.ParameterInput, class to use for specifying input of cls.
+    """
+    specs = super(AutoCorrelation, cls).getInputSpecification()
+    specs.name = 'AutoCorrelation'
+    specs.description = """AutoCorrelation TimeSeriesAnalysis algorithm."""
```
PaulTalbot-INL, line 37: This should probably be expanded to explain how it does what it does and why one might use it, a la user manual entry. You can copy the contents from the UserManual directly if you want.

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+    specs.addSub(InputData.parameterInputFactory(
+      'nlags',
+      contentType=InputTypes.IntegerType,
+      descr=""" Number of lags to return autocorrelation for. """
+    ))
+    specs.addSub(InputData.parameterInputFactory(
+      'alpha',
```
PaulTalbot-INL, line 44: can we rename this to something more user friendly than algorithm friendly? Maybe "confidence" or something like that?

I don't mind calling it "alpha" internally as long as user-facing parts (input in output) use a more readable term.

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+
+  def __init__(self, *args, **kwargs):
+    """
+      A constructor that will appropriately intialize a time-series analysis object
+      @ In, args, list, an arbitrary list of positional values
+      @ In, kwargs, dict, an arbitrary dictionary of keywords and values
+      @ Out, None
+    """
+    # general infrastructure
+    super().__init__(self, *args, **kwargs)
```
PaulTalbot-INL, line 60: do you need "self" in a `super()` call?

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+      This function utilizes the Discrete Wavelet Transform to
+      characterize a time-dependent series of data.
+
+      @ In, signal, np.ndarray, time series with dims [time, target]
+      @ In, pivot, np.1darray, time-like parameter values
+      @ In, targets, list(str), names of targets in same order as signal
+      @ In, settings, dict, additional settings specific to this algorithm
+      @ Out, params, dict, characteristic parameters
+    """
+    # TODO extend to continuous wavelet transform
```
PaulTalbot-INL, line 86: is this comment accurate?

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+    ## transformation as it assumed/required that each element in the
+    ## time-dependent series is independent, uniquely indexed and
+    ## sorted in time.
+    nlags = settings['nlags']
+    alpha = settings['alpha']
+    params = {target: {'results': {}} for target in targets}
+
+    for i, target in enumerate(targets):
+      results = params[target]['results']
+      results['signal1'], results['signal2'] = sm.acf(signal[:, i], nlags=nlags, alpha=alpha)
```
PaulTalbot-INL, line 104: I can't tell what `signal1` and `signal2` are by looking at them; can we name them something easier to decipher?

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+
+    return params
+
+
+
+  def generate(self, params, pivot, settings):
+    signal = 0
+    return signal
```
PaulTalbot-INL, line 112: missing docstrings; also I do not like this method implementation ... maybe it shouldn't be a generator.

framework/TSA/AutoCorrelation.py
```diff
@@ -0,0 +1,126 @@
...
+      Allows the engine to put whatever it wants into an XML to print to file.
+      @ In, writeTo, xmlUtils.StaticXmlElement, entity to write to
+      @ In, params, dict, trained parameters as from self.characterize
+      @ Out, None
+    """
+    for target, info in params.items():
+      base = xmlUtils.newNode(target)
+      writeTo.append(base)
+      for name, value in info['results'].items():
+        base.append(xmlUtils.newNode(name, text=','.join([str(v) for v in value])))
```
PaulTalbot-INL, line 126: is this the default implementation? If so it doesn't need to be repeated here.

framework/TSA/Factory.py
```diff
@@ -21,9 +21,12 @@
...
 from .Wavelet import Wavelet
+from .AutoCorrelation import AutoCorrelation
+from .PolynomialRegression import PolynomialRegression
 
 factory = EntityFactory('TimeSeriesAnalyzer')
 # TODO map lower case to upper case, because of silly ROM namespace problems
 aliases = {'Fourier': 'fourier',
-           'ARMA': 'arma'}
+           'ARMA': 'arma',
+           'Wavelet': 'wavelet'}
```
PaulTalbot-INL, line 31: what about ACF?

framework/TSA/PolynomialRegression.py
```diff
@@ -53,15 +53,15 @@ def __init__(self, *args, **kwargs):
       @ Out, None
     """
     # general infrastructure
-    TimeSeriesAnalyzer.__init__(self, *args, **kwargs)
+    super().__init__(self, *args, **kwargs)
```
PaulTalbot-INL, line 56: "self" in `super()` call?

framework/TSA/TimeSeriesAnalyzer.py
```diff
@@ -31,13 +31,30 @@ class TimeSeriesAnalyzer(utils.metaclass_insert(abc.ABCMeta, object)):
...
   @classmethod
   def getInputSpecification(cls):
     """
-      Method to get a reference to a class that specifies the input data for
-      class cls.
-      @ Out, inputSpecification, InputData.ParameterInput, class to use for
-        specifying input of cls.
+      Method to get a reference to a class that specifies the input data for class cls.
+
+      @ Out, specs, InputData.ParameterInput, class to use for specifying input of cls.
```
PaulTalbot-INL, line 57: I think this is missing the `@ In, None`

framework/TSA/Wavelet.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, *args, **kwargs):
       @ Out, None
     """
     # general infrastructure
-    TimeSeriesAnalyzer.__init__(self, *args, **kwargs)
+    super().__init__(self, *args, **kwargs)
```
PaulTalbot-INL, line 80: self in super call (sorry if I am wrong about this)

tests/framework/ROM/TimeSeries/SyntheticHistory/TrainingData/AutoCorrelation_A.csv
```diff
@@ -0,0 +1,2 @@
+scaling,filename
+1,AutoCorrelation_A_0.csv
```
PaulTalbot-INL, line 2: I noticed this file is pretty long. If that's what it takes to effectively test this, and it's still a fast test, no problem. Otherwise, can we get by with 100 entries instead?

tests/framework/ROM/TimeSeries/SyntheticHistory/autocorrelation.xml
```diff
@@ -0,0 +1,98 @@
+<?xml version="1.0" ?>
+<Simulation verbosity="debug">
+  <TestInfo>
+    <name>framework/ROM/TimeSeries/SyntheticHistory.AutoCorrelation</name>
+    <author>rinelson456</author>
+    <created>2021-04-19</created>
+    <classesTested>SupervisedLearning.SyntheticHistory,TSA.AutoCorrelation</classesTested>
+    <description>
+      Tests the SyntheticHistory ROM using only the AutoCorrelation TimeSeriesAnalyzer algorithm.
```
PaulTalbot-INL, line 9: See, this bothers me a bit because a characterizer shouldn't work in a SyntheticHistory ROM. Maybe we need that postprocessor.

tests/framework/ROM/TimeSeries/SyntheticHistory/polynomial.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version="1.0" ?>
```
PaulTalbot-INL, line 1: is this file a test? If so it's missing the TestInfo block.

tests/framework/unit_tests/TSA/testWavelet.py
```diff
@@ -243,12 +243,17 @@ def createWavelet(targets, family):
 params = transform.characterize(signals, pivot, targets, settings)
 check = params['A']['results']
 
+checkTrue("wavelet can generate", transform.canGenerate())
+checkTrue("wavelet can characterize", transform.canCharacterize())
+
```
PaulTalbot-INL, line 248: what about ACF cannot generate, can characterize?

PaulTalbot-INL (changes requested):
a handful of comments; we can discuss this more, but I think characterizers need a postprocessor, not the SyntheticHistory, to characterize signals well.

framework/TSA/AutoCorrelation.py
dylanjm: I'm down for changing the name but these argument do correspond directly to the named arguments in the underlying `statsmodels.acf` function. Does this change your perspective at all?

tests/framework/ROM/TimeSeries/SyntheticHistory/autocorrelation.xml
dylanjm: Yes, this was a problem that @rinelson456 and I talked about. I think that a PP would work well in this situation.

framework/TSA/AutoCorrelation.py
PaulTalbot-INL: I definitely am fine with using "alpha" internally; it's been our experience that way too many things have an "alpha" in the libraries we leverage, and it is frustrating from a user perspective to understand what a series of English and Greek letters mean when working in an input/output file. The description helps; if it can also be intuitive based on the name, all the better in my mind.

framework/TSA/Wavelet.py
dylanjm: I think you used to need a self argument in Python2 but in Python 3 that may no longer be needed. Although, I may be not understanding the post fully. Here is a SO citing the change between 2 and 3. 

https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods

PaulTalbot-INL: yeah, I tried this out:
```python
class A:
  def __init__(self, a, b):
    self._a = a
    self._b = b

class B(A):
  def __init__(self, a, b, c):
    super().__init__(self, a, b)
    self._c = c

b = B(1, 2, 3)
```
fails with `TypeError: __init__() takes 3 positional arguments but 4 were given`

framework/TSA/AutoCorrelation.py
rinelson456: Correct, The generate function just returns zero in this case. However if I remove the generate function and the TimeSeriesGenerator from AutoCorrelation it will crash. Any suggestions around this or should I leave it as is?

framework/TSA/AutoCorrelation.py
rinelson456: I will change alpha to confidence.

framework/TSA/AutoCorrelation.py
rinelson456: This will be changed to have a better description

framework/TSA/AutoCorrelation.py
rinelson456: I don't think you need self. I probably just missed removing it after I copied the file from wavelet.py.

framework/TSA/AutoCorrelation.py
rinelson456: I see that Time series Characterizer and Generator are no longer required so this issue will be fixed.

PaulTalbot-INL: I think Characterizer should stay; Generator can go though.

PaulTalbot-INL: @dylanjm and I were talking, I think we need a different tool than the SyntheticHistory generator to capture results for characterization instead of generation. I think he wants to pursue a PostProcessor for this.

framework/TSA/AutoCorrelation.py
rinelson456: Nope!

framework/TSA/AutoCorrelation.py
rinelson456: Generate has been removed

framework/TSA/AutoCorrelation.py
rinelson456: I am not familiar with what we would be using for signal1 or signal2 in this case, Do you have suggestions for different names?

framework/TSA/AutoCorrelation.py
rinelson456: Good to know!

framework/TSA/Factory.py
rinelson456: Interestingly autocorrelation does not need to be mapped like Wavelet or Arma,not sure if this is an issue or not.

framework/TSA/AutoCorrelation.py
PaulTalbot-INL: I honestly don't know what they are. Maybe @dylanjm has some ideas.

rinelson456: Dylan and I discussed it and we have changed signal1 to acf and signal2 to cofint.

framework/TSA/TimeSeriesAnalyzer.py
dylanjm: fixed

framework/TSA/Wavelet.py
dylanjm: fixed

tests/framework/ROM/TimeSeries/SyntheticHistory/polynomial.xml
dylanjm: fixed

dylanjm:
I think I will have to restructure this PR to get it merged, as I will need the modular structure to actually implement the LSTM TSA algo.

dylanjm:
@PaulTalbot-INL I am peeling off the autocorrelation commits in this PR into a separate PR in order to get the modular structure interface merged. Just as a heads up.

dylanjm:
@PaulTalbot-INL This is ready for a final review. 

QSUBS test is failing due to a missing "PIL" module which has nothing to do with any changes I've made.

PaulTalbot-INL (approved):
changes look good

PaulTalbot-INL:
This can be merged if the QSUB tests pass. The last attempt failed with a RAY parallel execution failure; I set them to re-try, as it is likely an execution fluke.

PR #1608, opened 2021-07-14
Cogljj/serialization exporter

joshua-cogliati-inl:
--------
Pull Request Description
--------
#1580

Adds serialization features.
This allows things like:
```<ExternalModel name="attenuate2" subType="pickledModel"/>```
which is a model that is loaded from a pickle.  It also allows external models to specify which variables are input and output variables.

----------------
----------------

### CODE REVIEW

File: tests/framework/Models/External/serialize_ext_model_and_use.xml [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+      <Input class="DataObjects" type="PointSet">placeholder</Input>
+      <Model class="Models" type="ExternalModel">attenuate2</Model>
+      <Sampler class="Samplers" type="Grid">grid</Sampler>
+      <Output class="DataObjects" type="PointSet">samples</Output>
+    </MultiRun>
+  </Steps>
+
+  <Models>
+    <ExternalModel ModuleToLoad="../AllMethods/attenuate" name="attenuate" subType="">
+      <variables>y1,y2,ans,fromInit,fromReadMoreXML,fromCNISelf,fromCNIDict</variables>
```
joshua-cogliati-inl, line 71: Hm, why doesn't this use the new input and output instead of variables?

framework/Models/Dummy.py
```diff
@@ -60,6 +60,17 @@ def specializeValidateDict(cls):
...
+  def copyModel(self, obj):
+    """
+      This method is aimed to copy the "obj" model in this instance
+      It is generally used for unpickling objects (models)
+      @ In, obj, instance, the instance of the object to copy from
+      @ Out, None
+    """
+    if obj.type != self.type:
+      self.raiseAnError(IOError,'Only objects of the same type can be copied! {} != {} !'.format(obj.type, self.type))
+    self.__dict__.update(obj.__dict__)
```
PaulTalbot-INL, line 72: This seems like a potentially painful avoidance of the setstate method. Maybe we should call a setstate method here that can be overloaded as needed?

framework/Models/ExternalModel.py
```diff
@@ -151,7 +165,9 @@ def localInputAndChecks(self,xmlNode):
       self.ModuleToLoad = paramInput.parameterValues['ModuleToLoad']
       moduleToLoadString, self.ModuleToLoad = utils.identifyIfExternalModelExists(self, self.ModuleToLoad, self.workingDir)
       # load the external module and point it to self.sim
-      self.sim = utils.importFromPath(moduleToLoadString, self.messageHandler.getDesiredVerbosity(self)>1)
+      self.sim = utils.importFromPath(moduleToLoadString,self.messageHandler.getDesiredVerbosity(self)>1)
+    elif paramInput.parameterValues['subType'].strip() == 'pickledModel':
+      self.pickled, self.constructed = True,  False
```
PaulTalbot-INL, line 170: multiple defs in one line is frowned on by the RAVAEN guidelines

framework/Models/ExternalModel.py
```diff
@@ -164,34 +180,51 @@ def localInputAndChecks(self,xmlNode):
       self.sim = ExternalModel.plugins.returnPlugin("ExternalModel",paramInput.parameterValues['subType'],self)
     else:
       self.raiseAnError(IOError,'"ModuleToLoad" attribute or "subType" not provided for Model "ExternalModel" named "'+self.name+'"!')
+    if not self.pickled:
+      # check if there are variables and, in case, load them
+      for child in paramInput.subparts:
+        if child.getName() == 'variables':
+          self.raiseAWarning(DeprecationWarning ,'"variables" node inputted but has been depreciated!  Please list variables in the "inputs" and "outputs" nodes instead.  This Warning will result in an error in RAVEN 3.0!')
+          if len(child.parameterValues) > 0:
+            self.raiseAnError(IOError,'the block '+child.getName()+' named '+child.value+' should not have attributes!!!!!')
```
PaulTalbot-INL, line 189: shouldn't this be an input params check?

framework/Models/ExternalModel.py
```diff
@@ -226,7 +259,10 @@ def _externalRun(self, Input, modelVariables):
       if key in self.modelVariableType.keys() or key in additionalKeys:
         InputDict[key] = Input[key]
 
-    self.sim.run(externalSelf, InputDict)
+    if 'current_time' in Input:
```
PaulTalbot-INL, line 262: this seems like a very implicit check; is there not a way to know ahead whether we're running in co-solve or independent solve mode? I assume that's what this means.

framework/Models/ROM.py
```diff
@@ -1428,6 +1433,19 @@ def provideExpectedMetaKeys(self):
...
 
+  def copyModel(self, obj):
+    """
+      This method is aimed to copy the "obj" model in this instance
+      It is generally used for unpickling objects (models)
+      @ In, obj, instance, the instance of the object to copy from
+      @ Out, None
+    """
+    # save reseeding parameters from pickledROM
+    loadSettings = self.initializationOptionDict
```
PaulTalbot-INL, line 1444: @wangcj05 this will cause problems with the new ROM API, possibly; can you check what issues this might raise?

framework/Steps/IOStep.py
```diff
@@ -118,10 +118,10 @@ def _localInitializeStep(self,inDictionary):
                                                        okay = 'Database',
                                                        received = inDictionary['Output'][i].type))
       # from ROM model to ...
-      elif isinstance(inDictionary['Input'][i], Models.ROM):
+      elif isinstance(inDictionary['Input'][i], Models.ROM) or isinstance(inDictionary['Input'][i], Models.ExternalModel):
```
PaulTalbot-INL, line 121: this can be changed to
```python
elif isinstance(inDictionary['Input'][i], (Models.ROM, Models.ExternalModel))
```
right? It would be easier to read and maintain the acceptable types this way.

framework/Steps/IOStep.py
```diff
@@ -134,23 +134,24 @@ def _localInitializeStep(self,inDictionary):
       # from File to ...
       elif isinstance(inDictionary['Input'][i],Files.File):
         # ... ROM
-        if isinstance(outputs[i],Models.ROM):
-          self.actionType.append('FILES-ROM')
+        if isinstance(outputs[i],Models.ROM) or isinstance(outputs[i],Models.ExternalModel):
```
PaulTalbot-INL, line 137: as above

framework/Steps/IOStep.py
```diff
@@ -207,44 +208,41 @@ def _localTakeAstepRun(self,inDictionary):
...
-        # print('CHECKING SIZE OF', unpickledObj)
-        # target = unpickledObj# .supervisedEngine.supervisedContainer[0]._macroSteps[2025]._roms[0]
-        # print('CHECKING SIZES')
-        # from utils.Debugging import checkSizesWalk
-        # checkSizesWalk(target, 1, str(type(target)), tol=2e4)
-        # print('*'*80)
-        # crashme
-        ## /DEBUGG
-        if not isinstance(unpickledObj,Models.ROM):
+        if not isinstance(unpickledObj,Models.ROM) and not isinstance(unpickledObj,Models.ExternalModel):
```
PaulTalbot-INL, line 228: tuple type check again

PaulTalbot-INL (commented):
A few comments to consider.

framework/Steps/IOStep.py
joshua-cogliati-inl: I agree, fixed (and the other two).

framework/Models/ExternalModel.py
joshua-cogliati-inl: It looks like it has just been copied from older code without realizing it.  Removing (since variables is already takes no subs).

framework/Models/Dummy.py
joshua-cogliati-inl: I am not sure what to do about this.  Since __setstate__ is private, it will not be inherited, so I think we might then need to implement it in every Model subclass if we use it instead of something like copyModel. (That said even if we leave it, ```copyModel``` should probably be ```_copyModel``` so people think twice about using it.)

PaulTalbot-INL: is `__setstate__` private? I think we're using inheritance in the SVL for get and set state currently, if I understand right.

framework/Models/ExternalModel.py
PaulTalbot-INL: ... is this even part of this PR? This looks like it belongs with time-step cross-model communication.

framework/Models/Dummy.py
joshua-cogliati-inl: How about changing it to this:
```python
    try:
      #If __getstate__ and __setstate__ available, use them
      self.__setstate__(obj.__getstate__())
    except AttributeError:
      #Otherwise just use object's dictionary
      self.__dict__.update(obj.__dict__)
```
So for classes where ```__setstate__``` and ```__getstate__``` are defined, it will use them, but otherwise it can just use the dictionary.

joshua-cogliati-inl: Updated docstring, renamed to _copyModel, and added setstate/getstate use.

PaulTalbot-INL:
Note that this PR (as of this writing) does not fully deprecate the `<variables>` node for ExternalModels, so a conversion script is not included. 

However, it does introduce this deprecation that should be finalized at some point.

PaulTalbot-INL (approved):

PaulTalbot-INL:
All comments have been addressed and checklist is complete. Merging.

PR #1658, opened 2021-09-16
DataSet from file as ROM training input

PaulTalbot-INL:
--------
Pull Request Description
--------
Addresses #1657, but not in devel (PR targets development branch)

Allows a DataSet to be loaded from file and used as a training input to a ROM.

## NOTE 

This PR targets a development branch, not devel.


aalfonsi (approved):
should this be merged before the ROM restructuring one?

wangcj05:
This PR is target for new ROM API branch. I think we can either merge it into ROM API branch or merge it into devel when ROM API got merged. @aalfonsi

wangcj05:
@PaulTalbot-INL I suggest we keep it as independent PR. First reason is the feature branch is already reviewed and is ready to be merged. I'm sorry about some delay since we want to merged it after our release. Second, I would like to keep the commit history about this branch.

Issue #1752, opened 2022-01-25
[TASK] Request metrics/statistics from BasicStatistics when using EconomicRatio


dgarrett622:
--------
Issue Description
--------

**Is your feature request related to a problem? Please describe.**
When requesting Value at Risk from EconomicRatio, it would be preferable to also be able to request additional metrics (including standard errors) on the NPV. The way to do this with the current implementation is to run another step with BasicStatistics.  

**Describe the solution you'd like**
Request BasicStatistics metrics from EconomicRatio directly. This would remove the need for a second postprocessing step for workflows where value at risk and metrics on NPV are requested.

**Describe alternatives you've considered**
EconomicRatio already inherits from BasicStatistics. The simplest solution seems to be to modify the methods in EconomicRatio to enable the metrics from BasicStatistics to pass through to EconomicRatio. From the user perspective, nothing changes for BasicStatistics and EconomicRatio will have enhanced features.

----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [ ] 1. Is it tagged with a type: defect or task?
- [ ] 2. Is it tagged with a priority: critical, normal or minor?
- [ ] 3. If it will impact requirements or requirements tests, is it tagged with requirements?
- [ ] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.
- [ ] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [ ] 1. If the issue is a defect, is the defect fixed?
- [ ] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [ ] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [ ] 4. If the issue is a defect, does it impact the latest release branch? If yes, is there any issue tagged with release (create if needed)?
- [ ] 5. If the issue is being closed without a pull request, has an explanation of why it is being closed been provided?

PaulTalbot-INL:
Approved to close via #1763.

PR #1763, opened 2022-02-03
EconomicRatio PostProcessor Enhancements

dgarrett622:
--------
Pull Request Description
--------
#1752 

- Metrics from BasicStatistics are requestable using EconomicRatio alone
- Standard error for percentile (in BasicStatistics) implemented including addition of percent to percentile standard error variable name for clarity - e.g., `percentile_10_ste_x`
- Standard error for value at risk (in EconomicRatio) implemented
- Added threshold to variable names valueAtRisk (including standard error), expectedShortfall, gainLossRatio, and sortinoRatio for clarity (EconomicRatio) - e.g., `VaR_0.05_NPV`, `VaR_0.05_ste_NPV`, `es_0.10_NPV`, `gainLoss_zero_NPV`, `sort_median_NPV`

----------------
----------------

### DIFF
--- a/framework/Models/PostProcessors/BasicStatistics.py
+++ b/framework/Models/PostProcessors/BasicStatistics.py
@@ -23,7 +23,7 @@ import copy
 from collections import OrderedDict, defaultdict
 import six
 import xarray as xr
-
+import scipy.stats as stats
 #External Modules End-----------------------------------------------------------
 
 #Internal Modules---------------------------------------------------------------
@@ -67,7 +67,8 @@ class BasicStatistics(PostProcessorInterface):
                 'variance_ste',
                 'sigma_ste',
                 'skewness_ste',
-                'kurtosis_ste']
+                'kurtosis_ste',
+                'percentile_ste']
 
   @classmethod
   def getInputSpecification(cls):
@@ -270,12 +271,21 @@ class BasicStatistics(PostProcessorInterface):
         for info in infos:
           prefix = info['prefix']
           for target in info['targets']:
-            metaVar = prefix + '_ste_' + target if not self.outputDataset else metric + '_ste'
-            metaDim = inputObj.getDimensions(target)
-            if len(metaDim[target]) == 0:
-              inputMetaKeys.append(metaVar)
+            if metric == 'percentile':
+              for strPercent in info['strPercent']:
+                metaVar = prefix + '_' + strPercent + '_ste_' + target if not self.outputDataset else metric + '_' + strPercent + '_ste'
+                metaDim = inputObj.getDimensions(target)
+                if len(metaDim[target]) == 0:
+                  inputMetaKeys.append(metaVar)
+                else:
+                  outputMetaKeys.append(metaVar)
             else:
-              outputMetaKeys.append(metaVar)
+              metaVar = prefix + '_ste_' + target if not self.outputDataset else metric + '_ste'
+              metaDim = inputObj.getDimensions(target)
+              if len(metaDim[target]) == 0:
+                inputMetaKeys.append(metaVar)
+              else:
+                outputMetaKeys.append(metaVar)
     metaParams = {}
     if not self.outputDataset:
       if len(outputMetaKeys) > 0:
@@ -591,7 +601,6 @@ class BasicStatistics(PostProcessorInterface):
     except IndexError:
       result = sortedWeightsAndPoints[indexL,1]
     return result
-
 
   def __runLocal(self, inputData):
     """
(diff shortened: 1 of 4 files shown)


dgarrett622:
I think tests will need to be modified or added for BasicStatistics, but I don't know which ones. I did not find tests for EconomicRatio, so suggestions on good tests to write are appreciated.

framework/Models/PostProcessors/BasicStatistics.py
```diff
@@ -290,10 +301,11 @@ def initialize(self, runInfo, inputs, initDict):
     metaKeys = inputMetaKeys + outputMetaKeys
     self.addMetaKeys(metaKeys,metaParams)
 
-  def _handleInput(self, paramInput):
+  def _handleInput(self, paramInput, childVals=[]):
```
PaulTalbot-INL, line 304: having an empty list as an optional argument in Python has unexpected consequences as a mutable object, see e.g. https://docs.python-guide.org/writing/gotchas/. Better off setting the default to `None` then checking `if None: val = []` in the beginning of the method.

framework/Models/PostProcessors/BasicStatistics.py
```diff
@@ -957,6 +969,40 @@ def __runLocal(self, inputData):
         percentileSet = percentileSet.rename({'quantile':'percent'})
       calculations[metric] = percentileSet
 
+      # because percentile is different, calculate standard error here
```
PaulTalbot-INL, line 972: if I recall correctly the document source for the ste of all these metrics is already cited in the code, right? If not, then we should note the source here.

framework/Models/PostProcessors/EconomicRatio.py
```diff
@@ -51,81 +56,26 @@ class cls.
...
-      for target in self.parameters['targets']:
-        pbName = 'ProbabilityWeight-' + target
-        if pbName in metaVars:
-          pbWeights[target] = dataSet[pbName]/dataSet[pbName].sum()
-        elif self.pbPresent:
-          pbWeights[target] = self.realizationWeight['ProbabilityWeight']
-    else:
-      self.raiseAWarning('EconomicRatio postprocessor did not detect ProbabilityWeights! Assuming unit weights instead...')
-    return inputDataset, pbWeights
+  def __init__(self):
```
PaulTalbot-INL, line 76: this constructor needs docstrings

framework/Models/PostProcessors/EconomicRatio.py
```diff
@@ -173,98 +162,118 @@ def _handleInput(self, paramInput):
       @ In, paramInput, ParameterInput, the already parsed input.
       @ Out, None
     """
-    self.toDo = {}
+
+    # handle scalarVals and vectorVals from BasicStatistics
+    super()._handleInput(paramInput, childVals=self.tealVals)
+
+    # now handle tealVals
```
PaulTalbot-INL, line 169: what are tealVals? They don't have to come from teal right?

PaulTalbot-INL: is that something that Jia/Andrea put in the original postprocessor that is just a misnomer name?

PaulTalbot-INL (changes requested):
This looks great! Nice implementation of the percentile STE and callback to BasicStatistics.

A couple comments and one change suggestion to review.

PaulTalbot-INL:
> I think tests will need to be modified

You can find the BasicStatistics with standard error tests in `raven/tests/framework/PostProcessors/BasicStatistics`, see for example `ste_mc.xml`. You should be able to add percentile STE to that test, probably.

For EconomicRatio, I sent an email seeing if anyone knows where one is; it might need to be created new.

framework/Models/PostProcessors/EconomicRatio.py
dgarrett622: tealVals were in the original code, they contain the metrics unique to EconomicRatio. They don't have to come from teal at all. I could rename them to something like "econVals" that might be more descriptive.

PaulTalbot-INL: not a required change, but I do think that would be vastly more informative, if you have other changes to make as well.

framework/Models/PostProcessors/BasicStatistics.py
dgarrett622: Changed default to None in commit 199977a

framework/Models/PostProcessors/BasicStatistics.py
dgarrett622: The reference was actually missing, I added it to both BasicStatistics and EconomicRatio in commit 063b945.

framework/Models/PostProcessors/EconomicRatio.py
dgarrett622: Oops, missed the docstring, added in commit 8f0fe30.

framework/Models/PostProcessors/EconomicRatio.py
dgarrett622: Easy find and replace task, I changed tealVals to econVals in commit 9338dde.

PaulTalbot-INL:
Code changes are good, we just need a test to cover EconomicRatio.

I note that several tests using BasicStatistics are failing due to the added percentile STE column, so we'll have to regold those after checking they're the same otherwise.

PaulTalbot-INL (approved):

PaulTalbot-INL:
Notification to plugin developers should be sent noting that use of the BasicStatistics is producing more information than it used to, which may fail some tests using the output from that postprocessor.

dgarrett622:
@wangcj05 do we want to send an email to the user group about the new percentile standard error?

framework/Models/PostProcessors/BasicStatistics.py
```diff
@@ -957,6 +988,46 @@ def __runLocal(self, inputData):
...
+        else:
+          calcPercentiles = calculatedPercentiles[target]
+          if targDa.values.min() == targDa.values.max():
+            # distribution is a delta function, so no KDE construction
+            percentileSte = list(np.zeros(calcPercentiles.shape))
+          else:
+            # get KDE
+            kde = stats.gaussian_kde(targDa.values, weights=targWeight)
+            factor = np.sqrt(np.array(percent)*(1.0 - np.array(percent))/en)
+            percentileSte = list(factor/kde(calcPercentiles.values))
```
wangcj05, line 1026: @dgarrett622 I have a question regarding the calculation of standard error for percentile, from the reference (see the following snapshot)
image

The standard normal distribution is used instead of KDE of real data, could you double check with the reference, and let me know what do you think?

dgarrett622: @wangcj05 That reference gives an equation for the asymptotic standard error estimate of a percentile assuming the random variable belongs to a normal distribution. There is a more general form of this asymptotic estimate that can be found in various references such as:

[http://www.medicine.mcgill.ca/epidemiology/hanley/bios601/DescriptiveStatistics/Var(percentile).pdf](http://www.medicine.mcgill.ca/epidemiology/hanley/bios601/DescriptiveStatistics/Var(percentile).pdf)

Here is a snapshot of the more general asymptotic estimate from the reference above which I implemented:
percentile_ste

The KDE tries to get the real distribution from the data without making the assumption that the data are normally distributed. The percentile, its standard error, and the KDE perform better with more samples, but that was always going to be a limitation when requesting a percentile.

dgarrett622: @wangcj05 The two equations are not equivalent. Using a Gaussian kernel is not the same as fitting a Gaussian to the data. The Gaussian kernel is like smoothing or interpolating to fit a distribution to the real data. Here is an example of using a Gaussian kernel to find a distribution that is definitely not Gaussian:

[https://gsalvatovallverdu.gitlab.io/python/kernel_density_estimation/](https://gsalvatovallverdu.gitlab.io/python/kernel_density_estimation/)

With the distribution fit to the data becoming:
gaussian_kde

The number of samples is an issue here, both in accuracy and computational cost. Determining the percentile itself requires many samples, especially if the requested percentile is in the tail of the distribution. The more samples you take, the more time it takes to compute the percentile since we do sorting on the data itself. The limitation in the paper applies to the calculation of the percentile itself and the standard error calculation, whether the standard normal formulation or the KDE method is used since it is based on number of samples. However, the standard normal formulation assumes that the random variable is distributed normally while the KDE method makes no assumption about the distribution.

An alternate method would be to use bootstrapped samples, but again there are concerns about computational time with that method.

Issue #1863, opened 2022-06-20
[TASK] Make RNG classes picklable


j-bryan:
--------
Issue Description
--------

**Is your feature request related to a problem? Please describe.**
Many models in RAVEN require using a random number generator, with an RNG engine in crow being used by default. Because these crow RNGs are being loaded with swig, they cannot be serialized with pickle. Currently, __getstate__ and __setstate__ methods must be implemented for each class which stores a swigged RNG instance so that the class can be serialized. This problem is currently causing SyntheticHistory ROM objects to not be picklable.

**Describe the solution you'd like**
It could be beneficial to wrap the swigged crow RNG instance so that it can be pickled, with the wrapper class reproducing the current RNG engine interface.

**Describe alternatives you've considered**
Addition of __getstate__/__setstate__ methods for the effected classes.

----------------
For Change Control Board: Issue Review
----------------
This review should occur before any development is performed as a response to this issue.
- [x] 1. Is it tagged with a type: defect or task?
- [x] 2. Is it tagged with a priority: critical, normal or minor?
- [x] 3. If it will impact requirements or requirements tests, is it tagged with requirements?
- [x] 4. If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.
- [x] 5. Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

-------
For Change Control Board: Issue Closure
-------
This review should occur when the issue is imminently going to be closed.
- [x] 1. If the issue is a defect, is the defect fixed?
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [x] 4. If the issue is a defect, does it impact the latest release branch? If yes, is there any issue tagged with release (create if needed)?
- [x] 5. If the issue is being closed without a pull request, has an explanation of why it is being closed been provided?

dylanjm:
@wangcj05 @PaulTalbot-INL You may have some insight or preferences for solving this issue. Please let us know your thoughts.

PaulTalbot-INL:
I think Jacob and I talked about this for a bit and ran a design past @wangcj05. I'd be interested in any ideas you have on minimizing the impact, but our basic idea was to make a light wrapper for the swig RNG engine instance that knows how to serialize itself safely.

dylanjm:
Approved to close via #1866

PR #1866, opened 2022-06-24
Makes random number generator class picklable

j-bryan:
--------
Pull Request Description
--------
#1863 

Crow RNG engines are now picklable. This was accomplished by adding a wrapper class for the crow RandomClass RNG class.

----------------
----------------

### DIFF
--- a/ravenframework/SupervisedLearning/ARMA.py
+++ b/ravenframework/SupervisedLearning/ARMA.py
@@ -21,7 +21,6 @@
 #External Modules------------------------------------------------------------------------------------
 import copy
 import collections
-from xml.dom.minidom import Attr
 from ..utils import importerUtils
 from ..utils.utils import findCrowModule
 statsmodels = importerUtils.importModuleLazy("statsmodels", globals())
@@ -467,8 +466,8 @@ class ARMA(SupervisedLearning):
     except AttributeError:  # catches where ARMA was pickled without saving the RNG
       self.setEngine(randomUtils.newRNG(), seed=None, count=rngCounts)
     else:
-      if type(self.randomEng) == findCrowModule('randomENG').RandomClass:
-        self.randomEng = randomUtils.RNG(self.randomEng, self.seed)  # wraps crow class if needed
+      if isinstance(self.randomEng, findCrowModule('randomENG').RandomClass):
+        self.randomEng = randomUtils.RNG(self.randomEng, self.seed)  # wraps raw crow RNG
 
     if self.reseedCopies:
       randd = np.random.randint(1, 2e9)
--- a/ravenframework/SupervisedLearning/ARMA.py
+++ b/ravenframework/SupervisedLearning/ARMA.py
@@ -21,7 +21,9 @@
 #External Modules------------------------------------------------------------------------------------
 import copy
 import collections
+from xml.dom.minidom import Attr
 from ..utils import importerUtils
+from ..utils.utils import findCrowModule
 statsmodels = importerUtils.importModuleLazy("statsmodels", globals())
 import numpy as np
 import functools
@@ -446,9 +448,9 @@ class ARMA(SupervisedLearning):
       @ Out, d, dict, stateful dictionary
     """
     d = super().__getstate__()
-    eng=d.pop("randomEng")
-    randCounts = eng.get_rng_state()
-    d['crow_rng_counts'] = randCounts
+    # eng=d.pop("randomEng")
+    # randCounts = eng.get_rng_state()
+    # d['crow_rng_counts'] = randCounts
     return d
 
   def __setstate__(self, d):
@@ -457,9 +459,17 @@ class ARMA(SupervisedLearning):
       @ In, d, dict, stateful dictionary
       @ Out, None
     """
-    rngCounts = d.pop('crow_rng_counts')
+    rngCounts = d.pop('crow_rng_counts', None)
     super().__setstate__(d)
-    self.setEngine(randomUtils.newRNG(), seed=None, count=rngCounts)
+
+    try:
+      self.randomEng
+    except AttributeError:  # catches where ARMA was pickled without saving the RNG
+      self.setEngine(randomUtils.newRNG(), seed=None, count=rngCounts)
+    else:
+      if type(self.randomEng) == findCrowModule('randomENG').RandomClass:
+        self.randomEng = randomUtils.RNG(self.randomEng, self.seed)  # wraps crow class if needed
+
     if self.reseedCopies:
       randd = np.random.randint(1, 2e9)
       self.reseed(randd)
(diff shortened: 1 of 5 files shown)


wangcj05:
Pretest failed: 
image

You can use script raven/developer_tools/delete_trailing_whitespace.sh yourFile to delete the white space.

ravenframework/SupervisedLearning/ARMA.py
```diff
@@ -446,9 +447,9 @@ def __getstate__(self):
       @ Out, d, dict, stateful dictionary
     """
     d = super().__getstate__()
-    eng=d.pop("randomEng")
-    randCounts = eng.get_rng_state()
-    d['crow_rng_counts'] = randCounts
+    # eng=d.pop("randomEng")
```
dylanjm, line 450: We should probably just remove the commented code.

ravenframework/SupervisedLearning/ARMA.py
```diff
@@ -2523,6 +2532,7 @@ def setEngine(self,eng,seed=None,count=None):
     if count is None:
       count=self.randomEng.get_rng_state()
     eng.forward_seed(count)
+    print('RNG State:', eng.get_rng_state())
```
dylanjm, line 2535: I think it would be good to either remove the print or use a `self.raiseAMessage()` instead.

dylanjm (changes requested):
Just some minor change requests. Everything else looks good to me.

j-bryan:
Thanks for the feedback @dylanjm and @wangcj05. The requested changes have been made.

dylanjm (approved):
Changes look good to me. This doesn't change any functionality and fixes an underlying bug in the SyntheticHistoryROM so no new tests or emails are required.

ravenframework/utils/randomUtils.py
```diff
@@ -85,11 +87,101 @@ def testSampling(self, n=1e5,engine=None):
...
+
+  def get_rng_seed(self):
+    """
+      Wrapper for RandomClass.get_rng_seed()
+      @ In, None
+      @ Out, int, RNG seed value
+    """
+    val = self._engine.get_rng_seed()  # returns int
+    self._seed = abs(int(val))
+    return self._seed
```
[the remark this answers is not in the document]
dylanjm: Completely missed those. @j-bryan You can make an additional PR for those changes.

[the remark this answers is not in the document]
j-bryan: @wangcj05 @dylanjm The function names in the wrapper class replicate the the crow RandomClass function names to keep the interface consistent. Changing those to camel case would likely require significant refactoring since those functions are used directly through RAVEN. Also, the remaining variable names in the wrapper class appear to follow the code standards, using camel case with one leading underscore in the case of protected variables. Am I misunderstanding that part of the style guide?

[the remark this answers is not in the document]
wangcj05: @j-bryan Here are what I found: 
image

I would suggest to change all of them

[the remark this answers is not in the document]
wangcj05: FYI, using one leading underscore is good to us.

ravenframework/SupervisedLearning/ARMA.py
```diff
@@ -457,9 +455,17 @@ def __setstate__(self, d):
...
     super().__setstate__(d)
-    self.setEngine(randomUtils.newRNG(), seed=None, count=rngCounts)
+
+    try:
+      self.randomEng
+    except AttributeError:  # catches where ARMA was pickled without saving the RNG
+      self.setEngine(randomUtils.newRNG(), seed=None, count=rngCounts)
+    else:
+      if isinstance(self.randomEng, findCrowModule('randomENG').RandomClass):
+        self.randomEng = randomUtils.RNG(self.randomEng, self.seed)  # wraps raw crow RNG
```
wangcj05, line 467: I guess you need to change randomUtils.RNG to randomUtils.CrowRNG, could you change it in your PR? @j-bryan

wangcj05: In addition, why the count is not settled in the self.randomEng? @j-bryan

j-bryan: For the `else` block in question, I think the best way forward would be to get rid of it entirely. I included it to handle an edge case where a raw crow RNG is found in `self.randomEng` when unpickling the ARMA ROM, but that shouldn't be possible. Eliminating that block shouldn't change the functionality and would be the cleanest solution. What are your thoughts @wangcj05?

I also found a mistake in the CrowRNG constructor docstring, so I'll include that correction and some minor cleanup of the constructor code in an update to the other PR.

j-bryan: To clarify on the handling of the count, it is handled in the CrowRNG wrapper. What's left in the __setstate__ method is so that existing pickled ROMs could still be loaded even if they didn't store `self.randomEng` when being pickled. Should I remove this, even if that'll cause existing ROM pickles to fail to load?

wangcj05: I think I agree with you, when self.randomEng exists, I think there is no need to set it again. 
Regarding the second question, I see your point, the count will be set when we load the engine. My question is mainly for above line where you try to reset the engine, but I think in this line, the count is not set yet. 
@j-bryan

ravenframework/utils/randomUtils.py
```diff
@@ -85,11 +87,101 @@ def testSampling(self, n=1e5,engine=None):
...
+
+  def __setstate__(self, d):
+    """
+      Set object instance state
+      @ In, d, dict, object state
+      @ Out, None
+    """
+    count = d.pop('count')
+    self.__dict__.update(d)
+    self._engine = findCrowModule('randomENG').RandomClass()  # reinstantiate RNG engine
```
wangcj05, line 132: It seems to me, you can make engine general enough to handle other cases, like numpy random engine. @j-bryan

j-bryan: The question seems to be about the scope of this RNG wrapper class. Originally, I had understood the issue as wanting to implement a thin wrapper of RandomClass in Crow so that could be pickled. The numpy random engine wasn't much of a concern since it those can be pickled already and the randomUtils functions already handle the differences between the numpy and crow engines. However, I could see it being beneficial to make a unified RNG class which would handle both the numpy and Crow RNGs in a uniform way.

ravenframework/SupervisedLearning/ARMA.py
j-bryan: @wangcj05 Yes, I think that line 467 should handle the count, if we don't just remove that section entirely. I believe the count is being handled correctly in the other cases.

wangcj05: @j-bryan I have two suggestions here:
1. remove the else check
2. make the CrowRNG to a more general wrapper for both Crow and numpy.

j-bryan: After talking point 2 over, it might be good to open a discussion on what the advantages a class-based RNG function interface would be over the current utility functions in randomUtils.py. @PaulTalbot-INL Perhaps you have some insights here?

Issue #1881, opened 2022-07-08
[DEFECT] too many threads modifying variable in DistributedMemoryRunner


joshua-cogliati-inl:
### Thank you for the defect report

- [X] I am using the latest version of `RAVEN`.
- [X] I have read the [Wiki](https://github.com/idaholab/raven/wiki).
- [ ] I have created a [minimum, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example)
that demonstrates the defect.

### Defect Description

Basically, in JobHandler terminateJobs which modifies the run list ( __running ) on a different thread than cleanJobQueue and fillJobQueue, both of which assume that only the loop thread can modify the run list.

In DistributedMemoryRunner, isDone checks self.thread.finished, but kill calls del self.thread
So if isDone is called after del self.thread, then self.thread.finished will error out.

```
Traceback (most recent call last):
  File "/home/fred/miniconda3/envs/raven_libraries_heron_newer_ray_and_h5py/li
b/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/home/fred/miniconda3/envs/raven_libraries_heron_newer_ray_and_h5py/li
b/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/projects/raven_db/fred/raven/ravenframework/JobHandler.py", line 442,
 in startLoop
    self.cleanJobQueue()
  File "/projects/raven_db/fred/raven/ravenframework/JobHandler.py", line 922, in cleanJobQueue
    if run is not None and run.isDone():
  File "/projects/raven_db/fred/raven/ravenframework/Runners/DistributedMemoryRunner.py", line 69, in isDone
    if self.thread is None:
AttributeError: 'DistributedMemoryRunner' object has no attribute 'thread'
```

### Steps to Reproduce

Run a long optimizer that uses terminateJobs.

### Expected Behavior

Not to crash.

### OS

Linux

### Dependency Manager

CONDA

### For Change Control Board: Issue Review

- [ ] Is it tagged with a type: defect or task?
- [ ] Is it tagged with a priority: critical, normal or minor?
- [ ] If it will impact requirements or requirements tests, is it tagged with requirements?
- [ ] If it is a defect, can it cause wrong results for users? If so an email needs to be sent to the users.
- [ ] Is a rationale provided? (Such as explaining why the improvement is needed or why current code is wrong.)

### For Change Control Board: Issue Closure

[...]

PR #1899, opened 2022-07-19
Combined lock

joshua-cogliati-inl:
--------
Pull Request Description
--------
Closes https://github.com/idaholab/raven/issues/1881

Adds a lock in DistributedMemoryRunner to prevent two threads stepping on each other.
Locks more in JobHandler cleanJobQueue because DistributedMemoryRunner isDone can't be called while kill is also being called.
Also makes a variable in DistributedMemoryRunner private because it really shouldn't be touched by other classes.

----------------
----------------


ravenframework/Runners/DistributedMemoryRunner.py
```diff
@@ -66,25 +72,26 @@ def isDone(self):
...
+          try:
+            ray.get(self.__func, timeout=waitTimeOut)
+            return True
+          except ray.exceptions.GetTimeoutError:
+            return False
+          except ray.exceptions.RayTaskError:
+            #The code gets this undocumented error, and
+            # I assume it means the task has unfixably died,
+            # and so is done
+            return True
```
PaulTalbot-INL, line 89: `True` means the job probably ended successfully? Is that the right return here? Do we believe the job probably finished, but the task didn't terminate as expected?

tests/framework/CodeInterfaceTests/RAVEN/optimizer.xml
```diff
@@ -0,0 +1,145 @@
...
+      Inner raven runs 100 MC samples, does basic stats, and returns the mean, sigma to the outer raven.
+    </description>
+    <analytic>
+      Each sample in this test converges towards the analytic solutions discussed in the "changing lower, upper bounds" section of
+      the "attenuate" analytic model documentation.
+    </analytic>
+    <revisions>
+      <revision author="talbpaul" date="2018-05-02">added "fully-correlated" variable for innerUpperBound as a test of the mechanic</revision>
+      <revision author="talbpaul" date="2018-11-06">modified inner run to use grid for speed, consistent results</revision>
+    </revisions>
```
PaulTalbot-INL, line 19: I don't think these revisions apply to this new test.

tests/framework/CodeInterfaceTests/RAVEN/optimizer.xml
```diff
@@ -0,0 +1,145 @@
+<Simulation verbosity="debug">
+  <TestInfo>
+    <name>framework/CodeInterfaceTests/RAVEN.Optimizer</name>
+    <author>talbpaul</author>
+    <created>2018-01-18</created>
```
PaulTalbot-INL, line 5: the author and creation date should probably be updated

tests/framework/CodeInterfaceTests/RAVEN/tests
```diff
@@ -5,6 +5,15 @@
...
    max_time = 500
  [../]
+ [./Optimizer]
+   type = 'RavenFramework'
+   input = 'optimizer.xml'
+   [./data]
+     type = OrderedCSV
+     output = 'Optimizer/opt_export_0.csv'
+   [../]
+   max_time = 500
```
PaulTalbot-INL, line 15: Does it really take this long? It doesn't look like it needed many optimization steps to find the solution on the boundary.

joshua-cogliati-inl: On the regression machines, the slowest I saw was 107 seconds, so I will update that.

ravenframework/Runners/DistributedMemoryRunner.py
joshua-cogliati-inl: Hm, well, isDone can either return true or false, and if it returns false, then JobHandler will think that it is still running, and so not remove it.  Setting returnCode = -1 might be the correct thing to do here.

tests/framework/CodeInterfaceTests/RAVEN/optimizer.xml
joshua-cogliati-inl: Agreed.

ravenframework/Runners/DistributedMemoryRunner.py
PaulTalbot-INL: Oh, okay, I misread, this is just telling if the guy is done nor not. `True` seems like a fair response.

joshua-cogliati-inl: (actually, this might be the cause of #1901 )

joshua-cogliati-inl: FYI: For future reference, the reason that returnCode = -1 needs to be set here as opposed to _collectRunnerResponse (which also uses ray.get) is that returnCode is checked before _collectRunnerResponse is called.

PaulTalbot-INL (approved):

PR #2189, opened 2023-09-28
Global TSA via ROMCollection

GabrielSoto-INL:
--------
Pull Request Description
--------
#2188 

New attribute for TSA Algorithms for "global" which defaults to False. If true, that algorithm is added to a new `_tsaGlobalAlgorithms` list otherwise they are added to the original `_tsaAlgorithms`. Calls to the normal `train` and `evaluate` within `TSAUser` have a new input to specify whether to use the local or global set of algorithms.

----------------
----------------


ravenframework/SupervisedLearning/SyntheticHistory.py
```diff
@@ -97,6 +98,65 @@ def __evaluateLocal__(self, featureVals):
...
+    for i,target in enumerate(targetNames):
+      trainingDict[target] = targetValues[:,:,i]
+    return settings, trainingDict
+
+  def finalizeGlobalRomSegmentEvaluation(self, settings, evaluation, weights, slicer):
+    """
+      Allows any global settings to be applied to the signal collected by the ROMCollection instance.
+      Note this is called on the GLOBAL templateROM from the ROMcollection, NOT on the LOCAL supspace segment ROMs!
+      @ In, evaluation, dict, {target: np.ndarray} evaluated full (global) signal from ROMCollection
+      TODO finish docs
```
dylanjm, line 139: looks like there is a few input params left to finish: `settings`, `weights`, `slicer`

ravenframework/SupervisedLearning/SyntheticHistory.py
```diff
@@ -97,6 +98,65 @@ def __evaluateLocal__(self, featureVals):
...
+      @ Out, evaluation, dict, {target: np.ndarray} adjusted global evaluation
+    """
+    if len(self._tsaGlobalAlgorithms)>0:
```
dylanjm, line 142: Should we raise a warning or error here if `len(self._tsaGlobalAlgorithms) == 0`?

ravenframework/SupervisedLearning/SyntheticHistory.py
```diff
@@ -97,6 +98,65 @@ def __evaluateLocal__(self, featureVals):
...
+      rlz = self.evaluateTSASequential(evalGlobal=True, evaluation=evaluation, slicer=slicer)
+      for key,val in rlz.items():
+        evaluation[key] = val
+    return evaluation
+
+  def finalizeLocalRomSegmentEvaluation(self,  settings, evaluation, globalPicker, localPicker=None):
```
dylanjm, line 148: Is this a required method to satisfy some abstract method definition? Seems odd to just return the input `evaluation`

ravenframework/TSA/TSAUser.py
```diff
@@ -95,6 +99,8 @@ def readTSAInput(self, spec):
     elif self.pivotParameterID not in self.target:
       # NOTE this assumes that every TSAUser is also an InputUser!
       raise IOError('TSA: The pivotParameter must be included in the target space.')
+    if len(self._tsaAlgorithms)==0:
+      print("No Segmenting algorithms were requested.")
```
dylanjm, line 103: Do we have access to the messageHandler through this class? Perhaps we should `self.raiseAWarning()` here instead of print.

ravenframework/TSA/TSAUser.py
```diff
@@ -211,20 +222,35 @@ def trainTSASequential(self, targetVals):
...
-  def evaluateTSASequential(self):
+  def evaluateTSASequential(self, evalGlobal=False, evaluation=None, slicer=None):
     """
       Evaluate TSA algorithms using a sequential linear superposition approach
-      @ In, None
-      @ Out, rlz, dict, realization dictionary of values for each target
+      @ In, evalGlobal, bool, are these algos trained on global signal?
+      @ In, evaluation, dict, realization dictionary of values for each target
+      @ In, slicer, list of slice, indexer for data range of this segment FROM GLOBAL SIGNAL
+      @ Out, evaluation, dict, realization dictionary of values for each target
```
dylanjm, line 231: I think this should be called `rlz` instead of `evaluation`

dylanjm (changes requested):
Some minor comments for consideration

ravenframework/SupervisedLearning/SyntheticHistory.py
GabrielSoto-INL: I had overloaded the method from `SupervisedLearning` because the inputs were different, but comparing with the `finalizeLocalRomSegmentEvaluation( )` method in the `SupervisedLearning.ARMA` it seems that the ARMA version just has an additional optional input that had not been updated in the parent version. 

I am removing this method from `SyntheticHistory.py`, updating the method in `SupervisedLearning.py` with the additional optional input it is missing so that it functions as the "abstract/do nothing" parent method (since it also just returns `evaluation`) for TSA

ravenframework/TSA/TSAUser.py
GabrielSoto-INL: not in `TSAUser`, but we have access in `SyntheticHistory.py`. will move the check to `handleInput` there

ravenframework/SupervisedLearning/SyntheticHistory.py
GabrielSoto-INL: this method is called in `ROMCollection` in `evaluate( )` and it seems like it was called regardless of whether a global or local algorithm was specified. the way I have it now, it just reports back the same evaluation if no global algorithms were specified so there shouldn't be an error as is

dylanjm (approved):

PR #2273, opened 2024-03-01
RavenFramework tests using pip package or pre-built executable

j-bryan:
--------
Pull Request Description
--------
#2272 

Adds the ability to run RavenFramework tests using either a pip-installed version of raven or a prebuilt executable of raven. Note that as currently implemented, the other types of raven tests (e.g. RavenPython) and the plugin tests will run using the development version of raven and its installed plugins, not the alternative installation.

----------------
----------------


tests/framework/unit_tests/CustomDrivers/basic_pip.xml
```diff
@@ -0,0 +1,93 @@
...
+    </IOStep>
+  </Steps>
+
+  <Files>
+    <Input name="inner_input" type="raven">../../../CodeInterfaceTests/RAVEN/Basic/inner_sample.xml</Input>
+  </Files>
+
+  <Models>
+    <Code name="raven" subType="RAVEN">
+        <executable>raven_framework</executable>
```
joshua-cogliati-inl, line 43: Could %RAVENFRAMEWORK% be used here, and then we could just use one test instead of duplicating tests?

rook/main.py
```diff
@@ -397,6 +407,21 @@ def term_handler(signum, _):
...
       print()
 
+  # The ravenframework installation may be of type "source", "pip", or "binary". Let the testers
+  # know which type of installation we are using.
+  if args.use_pip:
+    installType = 'pip'
+  elif args.use_binary:
+    # Specify an absolute path so we don't have to worry about where we are now vs where when
+    # we try to run the binary.
+    testers['RavenFramework'].set_binary_location(os.path.abspath(args.use_binary[0]))
```
joshua-cogliati-inl, line 417: This is kind of hackish, and will only work with RavenFramework.

rook/main.py
```diff
@@ -101,6 +101,12 @@ class NoColors:
 parser.add_argument('--python-command', dest='python_command',
                     help='command to run python')
 
+alt_framework_args = parser.add_mutually_exclusive_group(required=False)
+alt_framework_args.add_argument('--use-pip', dest='use_pip', action='store_true',
+                    help='use the pip-installed version of ravenframework')
+alt_framework_args.add_argument('--use-binary', nargs=1, dest='use_binary',
```
joshua-cogliati-inl, line 107: This doesn't seem to be tested currently (was it tested manually)?

joshua-cogliati-inl (changes requested):
I like where this patch is going and it will be useful. I think there are parts that can be simplified (if I am wrong feel free to explain what I missed).  Thanks.

rook/main.py
joshua-cogliati-inl: I think we should probably consider how to do the rook modifications better.

tests/framework/unit_tests/CustomDrivers/basic_pip.xml
j-bryan: I also don't like the duplication I've had to do with this test. The hurdle here is knowing how to run the inner RAVEN if the outer RAVEN was run with the PythonRaven interface. There's no clear way to recycle the outer run strategy for the inner in that case. That leaves us with manually specifying the way to run RAVEN for the inner, which is different for the different install types, hence the split test. All that said, if you have any ideas on how we can set this up without the duplication, it would certainly be preferable. I just can't think of how to reasonably set that up.

j-bryan:
The recent commit ([8c67f36](https://github.com/idaholab/raven/pull/2273/commits/8c67f363b99c595461d12af93a5d7ab4aafa07e5)) is a large rework of the capability with an emphasis on simplicity. The major changes are
- Renaming magic variable %RAVENFRAMEWORK% to %RAVENEXECUTABLE%, per the suggestion
- Handling interpretation of %RAVENEXECUTABLE% magic variable in `ravenframework/Models/Code.py` where the other magic variables are handled.
- Removal of the --use-pip and --use-binary command line options for Rook. Instead, a new command line option --tester-command can be used to specify the command to use to when running a certain tester. This removes RAVEN-specific instructions for Rook with the trade-off of a more difficult syntax for the developer running the tests. An example of using this option to set the RavenFramework tester command is `./run_tests --tester-command RavenFramework path/to/raven_framework`. In conjunction with the existing `--python-command` and `--command-prefix` options, this gives users full control over what a Tester runs as a command.
- Removal of the `skip_if_install_type` logic in RAVEN "tests" files. This was messy and a workaround for the lone python-raven-runs-raven test. However, I have not yet implemented a fix for this! This test is expected to fail if the magic variable %RAVENEXECUTABLE% is used to specify how to run raven in the inner raven for that test.

joshua-cogliati-inl (approved):
Reviewed and approve.

Issue #2567, opened 2026-03-26
[DEFECT] EnsembleModel permanently hangs at large batchSize:  lock convoy, failure cleanup deadlock, and missing fork() bypass extension point


chansoolee:
### Thank you for the defect report

- [x] I am using the latest version of `RAVEN`.
- [x] I have read the [Wiki](https://github.com/idaholab/raven/wiki).
- [ ] I have created a [minimum, reproducible example](https://stackoverflow.com/help/minimal-reproducible-example) that demonstrates the defect.

---

### Defect Description

`EnsembleModel` has three related bugs that cause permanent hangs or deadlocks at large `batchSize` (100+ concurrent evaluations in GA optimization). All three are silent — no error message is produced, making diagnosis difficult.

---

## Bug 1: Lock contention starvation (primary cause of hang)

**Location:** `EnsembleModel.py`, line 797

Each evaluation thread waits for its sub-model to finish via a busy-wait loop:

```python
while not jobHandler.isThisJobFinished(localIdentifier):
    time.sleep(1.e-3)
```

Every `isThisJobFinished()` call acquires `JobHandler.__queueLock`. With `N` concurrent evaluation threads (e.g., `batchSize=100`), this produces approximately **100,000 lock acquisitions per second** (`N × 1/0.001s`).

This starves `JobHandler.startLoop()`, which runs in a single polling thread and also requires `__queueLock` to execute `fillJobQueue()` and `cleanJobQueue()`. When the polling thread cannot acquire the lock, jobs that have already finished at the OS level never transition from `__running` to `__finished`, so `isThisJobFinished()` never returns `True` — causing a **permanent hang with no error message**.

This is a classic **lock convoy / priority inversion** problem. It does not manifest at small `batchSize` (e.g., 4–8) because lock contention is low enough for the polling thread to make progress.

---

## Bug 2: Failure cleanup blocks on never-submitted models

**Location:** `EnsembleModel.py`, line 811

When a sub-model in the pipeline fails, the cleanup code attempts to remove all other models from the job queue:

```python
for modelToRemove in list(set(self.orderList) - set([modelToExecute['Instance'].name])):
    jobHandler.getFinished(jobIdentifier=..., uniqueHandler=...)
```

[...]

PR #2322, opened 2024-05-18
Alfoa/hybrid model for batching and ensemble model

alfoa:
--------
Pull Request Description
--------
##### What issue does this change request address? (Use "#" before the issue to link it, i.e., #42.)
Closes #2321

##### What are the significant changes in functionality due to this change request?
The development here is addressing 2 issues:
- LogicalModel/HybridModel can now be used within an EnsembleModel 
- LogicalModel/HybridModel can now be used with Samplers/Optimizers that work using the ```batching``` strategy (e.g. GeneticAlgorithm).

In addition, an utility factory has been added in the JobHandler to collect and record Job Ids that have been used during the simulation. For now it is used as error checker but in the future can be used as base-class for identifier creation. 

Additional development has been carried out in the SERPENT interface to allow the user to specify a pre-calculation of the material volumes using SERPENT Montecarlo volume calculation routine.  A flag in SERPENT interface has been created to activate the stochastic material volume calculation via Serpent ``checkvolumes'' command. If True, the execution of  SERPENT is always preceded with a volume calculation and the user can use this file, including it in the SERPENT main input file.

----------------
For Change Control Board: Change Request Review
----------------
The following review must be completed by an authorized member of the Change Control Board.
- [x] 1. Review all computer code.
- [x] 2. If any changes occur to the input syntax, there must be an accompanying change to the user manual and xsd schema. If the input syntax change deprecates existing input files, a conversion script needs to be added (see Conversion Scripts).
- [x] 3. Make sure the Python code and commenting standards are respected (camelBack, etc.) - See on the [wiki](https://github.com/idaholab/raven/wiki/RAVEN-Code-Standards#python) for details.
- [x] 4. Automated Tests should pass, including run_tests, pylint, manual building and xsd tests. If there are changes to Simulation.py or JobHandler.py the qsub tests must pass.
- [x] 5. If significant functionality is added, there must  be tests added to check this. Tests should cover all possible options.  Multiple short tests are preferred over one large test. If new development on the internal JobHandler parallel system is performed, a cluster test must be added setting, in <RunInfo> XML block, the node ```<internalParallel>``` to True.
- [x] 6. If the change modifies or adds a requirement or a requirement based test case, the Change Control Board's Chair or designee also needs to approve the change.  The requirements and the requirements test shall be in sync.
- [x] 7. The merge request must reference an issue.  If the issue is closed, the issue close checklist shall be done.
- [x] 8. If an analytic test is changed/added is the the analytic documentation updated/added?
[...]

PR #2580, opened 2026-05-19
Chansoo/ensemble jobhandler event wait

chansoolee:
--------
Pull Request Description
--------
#2567 

This change stabilizes EnsembleModel job completion when using the threaded parallel strategy. Previously, each ensemble evaluation thread repeatedly polled `JobHandler.isThisJobFinished()`, which acquired the shared JobHandler queue lock. With many concurrent ensemble evaluations, this polling could create enough lock contention to prevent the JobHandler polling loop from promptly moving completed jobs into the finished queue, producing hangs even after subprocesses had completed.

The JobHandler now creates a per-job `threading.Event` when a job is submitted, sets that event when the job is moved to the finished queue, and exposes `getJobEvent()` so EnsembleModel can wait on job completion without repeatedly acquiring the queue lock. The previous polling path remains as a fallback if no event is available.

This PR also improves failed sub-model handling in EnsembleModel. Failed evaluations now return no evaluation and allow the existing failed-run handling path to process the failure, instead of raising immediately and potentially leaving nested ensemble jobs unclaimed. Cleanup is limited to models that were actually submitted before the failed model, avoiding waits on downstream models that were never launched.

Additionally, `MessageHandler.printWarnings()` now clears stored warnings after printing so warnings are not reprinted later by subsequent error handling. A focused JobHandler unit test was added to verify event creation, event signaling on completion, and event cleanup after finished-job collection.

  ##### Tests performed

The following checks were run in `/home/leec3/projects/raven_idaholab_issue2567`:

  - `python -m py_compile ravenframework/JobHandler.py ravenframework/MessageHandler.py ravenframework/Models/EnsembleModel.py tests/framework/unit_tests/JobHandler/testJobHandlerEvents.py`
  - `python -m py_compile ravenframework/CustomModes/PBSSimulationMode.py ravenframework/CustomModes/MPILegacySimulationMode.py`
  - `./run_tests --test-dir tests/framework/unit_tests/JobHandler`
  - `./run_tests --test-dir tests/framework/ensembleModelTests --re 'testEnsembleModelLinearThread|testEnsembleModelLinearExpectedFailure|testEnsembleModelWithCode' -j 3`
  - `git diff --check`

  Cluster/qsub validation was also run through Slurm:

  - Slurm job `1873408`
  - Command path: `./run_tests --raven --test-dir tests/cluster_tests/InternalParallel --re 'test_parallel_ensemble_linear' --run-types qsub -j 1`
  - Result: `COMPLETED`, exit code `0:0`
  - ROOK result: `PASSED: 1`, `FAILED: 0`
  - Final run_tests result: `Tested sets: RAVEN PluginAPI`, `... tests passed!`

----------------
----------------

### DIFF
--- a/ravenframework/Models/EnsembleModel.py
+++ b/ravenframework/Models/EnsembleModel.py
@@ -532,7 +532,10 @@ class EnsembleModel(Dummy):
     Input = self.createNewInput(myInput[0], samplerType, **kwargsToKeep)
 
     ## Unpack the specifics for this class, namely just the jobHandler
-    returnValue = (Input,self._externalRun(Input, jobHandler))
+    evaluation = self._externalRun(Input, jobHandler)
+    if evaluation is None:
+      return None
+    returnValue = (Input,evaluation)
     return returnValue
 
   def submit(self,myInput,samplerType,jobHandler,**kwargs):
@@ -716,10 +719,13 @@ class EnsembleModel(Dummy):
         ##if self.runInfoDict and 'Code' == self.modelsDictionary[modelIn]['Instance'].type:
         ##  inputKwargs[modelIn].update(self.runInfoDict)
 
-        retDict, gotOuts, evaluation = self.__advanceModel(identifier, self.modelsDictionary[modelIn],
-                                                        originalInput[modelIn], inputKwargs[modelIn],
-                                                        inRunTargetEvaluations[modelIn], samplerType,
-                                                        iterationCount, jobHandler)
+        evaluationInfo = self.__advanceModel(identifier, self.modelsDictionary[modelIn],
+                                             originalInput[modelIn], inputKwargs[modelIn],
+                                             inRunTargetEvaluations[modelIn], samplerType,
+                                             iterationCount, jobHandler)
+        if evaluationInfo is None:
+          return None
+        retDict, gotOuts, evaluation = evaluationInfo
 
         returnDict[modelIn] = retDict
         typeOutputs[modelCnt] = inRunTargetEvaluations[modelIn].type
@@ -781,6 +787,7 @@ class EnsembleModel(Dummy):
       suffix = f"{utils.returnIdSeparator()}{inputKwargs['batchRun']}"
     self.raiseADebug('Submitting model',modelToExecute['Instance'].name)
     localIdentifier = f"{modelToExecute['Instance'].name}{utils.returnIdSeparator()}{identifier}{suffix}"
+    excType, excValue, excTrace = RuntimeError, RuntimeError("Model returned no evaluation"), None
     if self.parallelStrategy == 1:
       # we evaluate the model directly
       try:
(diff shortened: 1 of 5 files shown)


ravenframework/Models/EnsembleModel.py
```diff
@@ -794,9 +801,46 @@ def __advanceModel(self, identifier, modelToExecute, origInputList, inputKwargs,
...
+        #   2. Acts as a safety net in case the Event is somehow missed
+        #      (should not happen in normal operation).
+        #
+        # Backward compatibility:
+        #   If getJobEvent() returns None (unexpected), we fall back to
+        #   the original polling loop so that non-standard JobHandler
+        #   subclasses or configurations still work.
+        jobEvent = jobHandler.getJobEvent(localIdentifier)
+        if jobEvent is not None:
+          while not jobEvent.wait(timeout=30.0):
```
wangcj05: Could you justify why use 30 here for timeout?

ravenframework/Models/EnsembleModel.py
```diff
@@ -794,9 +801,46 @@ def __advanceModel(self, identifier, modelToExecute, origInputList, inputKwargs,
...
+            pass
+        else:
+          # Fallback: original polling (should not be reached in normal use)
+          self.raiseAWarning(f'No Event found for job "{localIdentifier}", '
+                             f'falling back to polling-based wait')
+          while not jobHandler.isThisJobFinished(localIdentifier):
+            time.sleep(1.0)
```
wangcj05: could you justify why 1.0 is used here? Could this cause a issue when the model can run very quickly?

ravenframework/Models/EnsembleModel.py
```diff
@@ -808,23 +852,35 @@ def __advanceModel(self, identifier, modelToExecute, origInputList, inputKwargs,
...
-                                 uniqueHandler = f"{self.name}{identifier}{suffix}")
+        # the model failed — clean up only models that were submitted
+        # BEFORE the failed model (they may be running or finished).
+        # Models AFTER the failed one in orderList were never submitted,
+        # so calling getFinished for them would block indefinitely.
+        failedModelIdx = self.orderList.index(modelToExecute['Instance'].name)
+        for modelToRemove in self.orderList[:failedModelIdx]:
+          try:
+            jobHandler.getFinished(jobIdentifier = f"{modelToRemove}{utils.returnIdSeparator()}{identifier}{suffix}",
+                                   uniqueHandler = f"{self.name}{identifier}{suffix}")
```
wangcj05: I agree with the comments. Regarding the implementation, when job run in parallel, job execution time can vary, which means failed job may appear in the middle of submitted jobs. If this is the case, the clean up can fail. Could you verify it?

wangcj05 (changes requested):
@chansoolee I have few comments for you to consider.

ravenframework/Models/EnsembleModel.py
chansoolee: This fallback is only used if the per-job Event is unavailable, but using 1.0 seconds can add unnecessary latency for fast-running models. I changed the fallback interval to 0.1 seconds instead of restoring the original 1.e-3 interval, since the fallback still calls isThisJobFinished() and therefore still takes the JobHandler queue > lock. This keeps fallback latency bounded while avoiding the high-frequency polling pattern that caused the original contention problem.

ravenframework/Models/EnsembleModel.py
chansoolee: The 30 second timeout was arbitrary. 

The original intent was to avoid an indefinite wait without periodically returning control to Python, but the loop body did not perform any recovery, state check, or logging after timeout. 

In practice, it only woke up every 30 seconds and immediately waited again. I removed the timeout loop and now wait directly on the per-job Event with `jobEvent.wait()`. 

A normal model failure should still wake this wait: the Runner exits, `JobHandler.cleanJobQueue()` moves the job to the finished queue, and `event.set()` is called. The failure is then handled through the existing Runner/JobHandler return-code path.

ravenframework/Models/EnsembleModel.py
chansoolee: For `EnsembleModel`, the parallelism is across ensemble evaluations, limited by xml input named `batchSize`. Within a single ensemble evaluation, the sub-models in `orderList` are still advanced sequentially: `__advanceModel()` submits one sub-model, waits for that specific job to finish, collects it with `getFinished(localIdentifier, uniqueHandler=...)`, and only then moves to the next sub-model. 

So for one ensemble evaluation, when this failure branch is reached:
 - any previous sub-models in `orderList` have already been collected (not running),
 - the failed current sub-model has just been collected, (not running)
 - downstream sub-models have not been submitted yet. (not running or in the waitlist) 
 
Therefore the cleanup loop is not required for correctness. It does not remove any live downstream work in the current execution model, and for previous sub-models it can only try to collect jobs that were already collected earlier. Keeping the loop would make the code depend on `orderList` as a proxy for the set of submitted jobs, which is exactly the fragile assumption.

I removed the extra cleanup loop and let the failed sub-model return `None`, so the enclosing `EnsembleModel` evaluation follows the existing failed-run handling path used by the outer `JobHandler`/`MultiRun`.

chansoolee: After tracing the PR#2322, here's the summary.

`orderList` traversal is still sequential after #2322.

In `_externalRun` (lines 673–728), each sub-model goes through `__advanceModel,` which does `submit → wait (Event or polling) → getFinished` before the loop advances. Two `orderList` sub-models of the same ensemble evaluation are never in flight simultaneously, and the dependency ordering enforces this in principle as well.

Where #2322 placed parallelism, and why the removed loop wasn't on that path:
- `localJobHandler` (lines 566–572): inner queue for` parallelStrategy == 2`. Does not alter _externalRun's iteration.
- Batching of ensemble evaluations (line 578): each evaluation is a separate client job, isolated by `uniqueHandler = f"{self.name}{identifier}{suffix}"`. Failures in one evaluation cannot reach another's cleanup scope.
- HybridModel/LogicalModel as sub-models: their internal jobs are owned and reaped by themselves, not by `orderList,` so the removed loop never touched their identifiers.

What the removed loop actually did.

```
for modelToRemove in self.orderList[:failedModelIdx]:
    try:
        jobHandler.getFinished(jobIdentifier=..., uniqueHandler=...)
    except Exception:
        pass
```

Because `orderList` runs sequentially, every `modelToRemove` before `failedModelIdx` had already been collected by the per-iteration `getFinished` at line 840. The failed sub-model was just collected in the current iteration. Downstream sub-models were never submitted. The loop was re-querying already-reaped identifiers and silently swallowing the resulting exceptions, so it was not a guard against parallel-failure scenarios.

Verification: Re-ran the parallel/batched paths affected by #2322 on this branch — all pass:
- test_em_connection.xml (batchSize=2)
- test_ensemble_model_picard_parallel.xml (batchSize=2, Picard)
- test_ensemble_model_linear_internal_parallel_with_optimizer.xml (GradientDescent)

If a HybridModel/LogicalModel sub-model fails internally, draining its own internal queue is its own responsibility. The removed loop never covered that case, so this change does not regress it.

wangcj05 (approved):
changes are good.

wangcj05:
tests are good also.

PR #2583, opened 2026-06-23
PRLO modifications to Optimizer-related classes

### LINKED ISSUE #2587: [TASK] Update RAVEN with PRLO-related changes to optimizer
--------
Issue Description
--------

**Is your feature request related to a problem? Please describe.**
PRLO developments were conducted on a separate fork of RAVEN. Some of these additions are valuable to RAVEN, generally, and need to be folded back into the main fork.

**Describe the solution you'd like**
A merge-ready diff for migrating the PRLO-related changes and additions into RAVEN main fork needs to be prepared.

----------------
----------------
- [x] 1. Is it tagged with a type: defect or task?

-------
-------
- [x] 1. If the issue is a defect, is the defect fixed?
- [x] 2. If the issue is a defect, is the defect tested for in the regression test system?  (If not explain why not.)
- [x] 3. If the issue can impact users, has an email to the users group been written (the email should specify if the defect impacts stable or master)?
- [x] 4. If the issue is a defect, does it impact the latest release branch? If yes, is there any issue tagged with release (create if needed)?
- [x] 5. If the issue is being closed without a pull request, has an explanation of why it is being closed been provided?

W0lfShAd0w:
--------
Pull Request Description
--------
close #2587 

This PR has been reworked to create a merge-ready diff for migrating the PRLO-related changes and additions into the RAVEN devel branch.
This PR upstreams the optimizer robustness features the PRLO plugin needed so that PRLO-related RAVEN forks can be retired. The core additions are a checkpoint/restart system and exact-match submission deduplication for RavenSampled-based optimizers (GA, Bayesian, GradientDescent, SimulatedAnnealing), GA-specific z-score fitness normalization, a fix reverting the default penalty-scaling factor to 1.0 alongside generalizing penalty scaling to per-(objective, constraint) arrays, and new registerCrossover/registerMutator plugin hooks that let PRLO-specific GA operators live entirely in the plugin instead of RAVEN core.
This PR will close all PRLO-related RAVEN forks as active RAVEN versions.

----------------
----------------


ravenframework/Optimizers/RavenSampled.py
```diff
@@ -385,7 +779,40 @@ def finalizeSampler(self, failedRuns):
...
                         'Perhaps the Model failed?')
 
+    ## If any solution in the population has a higher fitness value than what is found in ._optPointHistory, add it to opt. NOTE: this only searches the most recent iteration of the optimizer.
     opt = self._optPointHistory[traj][-1][0]
+    fitnessVars = [var for var in list(opt) if 'fitness' in var.lower()]
+    self._solutionExport.asDataset() #empty _collector into _data
+    ## Search the rest of the population
+    if hasattr(self,"_sampledPopulationInfo"):
+      bestInPopulation = {}
+      # find best solns from the population
```
Jimmy-INL: @wangcj05 , @W0lfShAd0w , @JunyungKim What is this doing here? RavenSampled is not Genetic Algorithm, it is actually a base class that all optimizers inherit from. Why does it have fitness and other terms of GA? How will Gradient Descent, Simulated annealing, Bayesian Optimization act when it sees these variables? If the answer is: all of those were modified, this would be even worse! I must be missing something here.

ravenframework/Optimizers/mutators/mutators.py
```diff
@@ -51,13 +53,215 @@ def swapMutator(offSprings, distDict, **kwargs):
...
+    for single-cycle (Nth cycle) problems based on the calculationType in the prlodata file.
+    @ In, offSprings, xr.DataArray, children resulting from the crossover process
+    @ In, distDict, dict, dictionary containing distribution associated with each gene
+    @ In, kwargs, dict, see swapMutatorEQ / swapMutatorSingleCycle for full parameter list.
+    @ Out, children, xr.DataArray, the mutated chromosome, i.e., the child.
+
+    #!TODO(rollnk): Deprecated. Replaced by PRLO/src/Optimizers/mutators.py::swapMutatorSS.
+    #!              Remove this copy once the PRLO version is validated in production.
+  """
+  if not any("prlodata" in sublist for sublist in kwargs["files"]):
```
Jimmy-INL: Are the swapMutatorSS, swapMutatorEQ, swapMutatorSingleCycle PRLO specific? Didn't we agree that these will be only in PRLO? Can any other user of the RAVEN optimizer use them on any other optimization other than fuel reload? If the answer is no, why are they here? even `prlodata` is mentioned here, that cannot be the case.

ravenframework/Optimizers/crossOverOperators/crossovers.py
```diff
@@ -217,3 +256,229 @@ def uniformCrossoverMethod(parent1,parent2,crossoverProb):
       children2[pos] = parent2[pos]
 
   return children1,children2
+
+def uniformEQCrossoverMethod(parent1,parent2,crossoverProb,eqchecker):
+  """
+    Uniform crossover for equilibrium-cycle (EQ) PRLO shuffling schemes.
+    Two crossover passes are performed within each retry iteration:
+
+    Pass 1 — Fresh-batch (batch-1) crossover: positions where both parents carry
```
Jimmy-INL: Same here. This should be in PRLO

scripts/conversionScripts/convert_globalseed.py
```diff
@@ -37,33 +35,11 @@ def convert(tree,fileName=None):
...
   return tree
 
 if __name__=='__main__':
   import convert_utils
   import sys
   convert_utils.standardMain(sys.argv,convert)
   ## the following is provided as an example format for the expected content of the system arguments (sys.argv):
-  #sys.argv = ['/home/rollnk/LWRS-PRLO/raven_deployed/scripts/conversionScripts/convert_globalseed.py','--tests','--no-rewrite','--remove-comments','/home/rollnk/LWRS-PRLO/raven_deployed/tests/']
\ No newline at end of file
+  #sys.argv = ['/home/rollnk/LWRS-PRLO/raven_latest/scripts/conversionScripts/convert_globalseed.py','--tests','--no-rewrite']
```
Jimmy-INL: ???
This is a really bad example.
1. maybe '~' instead of the personal hardcoded `/home/rollnk`
2. This doesn't use `projects` as all raven docs and installation assumes and advise.
3. raven folder is called `raven_latest` which is the worst practice, this reflects that we are not using version control at all :). Unless this is trying to encourage users to use git worktrees in lieu of branches. Please generalize this or even remove it.

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: This is a good criticism. The concept here is easily generalizable: "population" and "history" should be synonymous. "Fitness" and "optimization goal" are also equivalent. For the time being, I'm using "self._objectiveVar" to represent the optimization goal, though I don't think that evaluates to "fitness" for GA. @Jimmy-INL @wangcj05 do we have an attribute that represents the "goal" value for the optimizer?

ravenframework/Optimizers/mutators/mutators.py
W0lfShAd0w: As noted in the TODO message in the docstring, these methods have been migrated to PRLO and the ones found here should not be used. They have been deprecated and will be deleted once it has been deemed safe to do so.

ravenframework/Optimizers/crossOverOperators/crossovers.py
W0lfShAd0w: As noted in the TODO message in the docstring, these methods have been migrated to PRLO and the ones found here should not be used. They have been deprecated and will be deleted once it has been deemed safe to do so.

ravenframework/Optimizers/fitness/fitness.py
```diff
@@ -130,9 +143,9 @@ def feasibleFirst(rlz, **kwargs):
           # if constraints are violated
           else:  # Penalize constraint violations
               fit = -a[i] * worstObj  # Start with the worst objective value
-              for constInd in range(g.data.shape[1]):
-                  violation = max(0, -g.data[ind, constInd])
-                  fit -= b[i] * violation
+              for constInd in range(constraintNum):
+                  violation = min(0, g.data[ind, constInd])
+                  fit += b[i][constInd] * violation #add the negative penalty
```
wangcj05, line 148: Please explain the reason why the sign changes here. from ``-'' to ``+''.

ravenframework/Optimizers/fitness/fitness.py
```diff
@@ -154,7 +167,7 @@ def logistic(rlz, **kwargs):
         scale, list of floats, scaling coefficient(s) for the objective function (default = 1.0 for each objective)
         shift, list of floats, coefficient(s) for shifting the objective value (default = 0.0 for each objective)
         constraintFunction, xr.DataArray, measuring the severity of the constraint violation
-        penalty, list of floats, penalties for constraint violations (default = 1.0 for each objective)
+        penalty, list of floats, penalties for constraint violations (default = 10.0 for each objective)
```
wangcj05, line 170: please justify the change of default values here.

ravenframework/Optimizers/BayesianOptimizer.py
```diff
@@ -296,16 +296,15 @@ def _submitRun(self, point, traj, step, moreInfo=None):
...
     """
     info = {}
     if moreInfo is not None:
       info.update(moreInfo)
     info.update({ 'traj': traj,
                   'step': step,
                   'batchSize': self.batch})
-    self.raiseADebug(f'Adding run to queue: {self.denormalizeData(point)} | {info}')
-    self._submissionQueue.append((point, info))
+    return self._queueSubmission(point, info)
```
wangcj05, line 307: maybe worth to keep the original raiseADebug message

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/mishraBirdConstrainedInvLin/opt_export_0.csv
```diff
@@ -149,104 +149,104 @@ x,y,ans,age,batchId,fitness,iteration,accepted,conv_AHDp
...
+-4.67545493615,-2.29571380042,-1.51116048449,0.0,4.0,1.51116048449,3.0,accepted,0.0
+-3.97213496593,-2.40681328355,-35.687869281,0.0,4.0,35.687869281,3.0,accepted,0.0
+-2.56422205189,-2.29571380042,-57.1483747952,0.0,4.0,57.1483747952,3.0,accepted,0.0
+-3.97213496593,-2.29571380042,-40.7741461662,0.0,4.0,40.7741461662,3.0,accepted,0.0
+-3.97213496593,-2.29571380042,-40.7741461662,0.0,4.0,40.7741461662,3.0,accepted,0.0
+-3.97213496593,-1.13468240542,-48.2949092726,0.0,4.0,48.2949092726,3.0,accepted,0.0
+-1.74566523911,-2.29571380042,-13.2752113996,0.0,4.0,13.2752113996,3.0,accepted,0.0
+-2.56422205189,-2.58188772786,-38.8650663852,0.0,4.0,38.8650663852,3.0,accepted,0.0
+-3.92580883902,-2.29571380042,-43.5077903642,0.0,4.0,43.5077903642,3.0,accepted,0.0
+-3.97213496593,-2.29571380042,-40.7741461662,0.0,4.0,40.7741461662,3.0,accepted,0.0
```
wangcj05, line 177: Could you explain why the age becomes 0.0 for all the trajectories?

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/rosenbrockDiskConstrained/opt_export_0.csv
```diff
@@ -99,1404 +99,1404 @@ x,y,ans,age,batchId,fitness,iteration,accepted,AHDp,conv_AHDp
...
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864154,29.6121144078,18.0,26.0,-29.6121144078,25.0,accepted,0.0488339074079,0.0
+0.367768831986,0.675738864
```
wangcj05, line 1496: same here, could you explain why the age is always 0.0

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/simionescuConstrainedInvLin/opt_export_0.csv
```diff
@@ -79,124 +79,124 @@ x,y,ans,age,batchId,fitness,iteration,accepted,AHDp,conv_AHDp
...
--0.501388060919,0.476866056381,-0.0239094947327,1.0,5.0,0.0239094947327,4.0,accepted,0.189159514705,0.0
--0.785396892935,0.544481988017,-0.0427634461648,1.0,5.0,0.0427634461648,4.0,accepted,0.189159514705,0.0
-0.476866056381,-0.501388060919,-0.0239094947327,1.0,5.0,0.0239094947327,4.0,accepted,0.189159514705,0.0
--0.785396892935,0.461104774548,-0.0362150257247,1.0,5.0,0.0271248242913,4.0,accepted,0.189159514705,0.0
--0.501388060919,0.461104774548,-0.0231192428791,1.0,5.0,0.0231192428791,4.0,accepted,0.189159514705,0.0
--0.601577741176,0.544481988017,-0.0327548244462,1.0,5.0,0.0327548244462,4.0,accepted,0.189159514705,0.0
--0.501388060919,0.461104774548,-0.0231192428791,1.0,5.0,0.0231192428791,4.0,accepted,0.189159514705,0.0
-0.461104774548,-0.601577741176,-0.0277390368718,1.0,5.0,0.0277390368718,4.0,accepted,0.189159514705,0.0
-0.735111348793,-0.871549073984,-0.0640685615316,0.0,5.0,0.0640685615316,4.0,final,0.189159514705,0.0
+0.461104774548,-0.501388060919,-0.0231192428791,0.0,3.0,0.0231192428791,2.0,accepted,0.346865937582,0.0
```
wangcj05, line 82: It seems the convergence becomes slower, could you explain?

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/unconstrained/ensemble_withGA_HDSM/opt_export.csv
```diff
@@ -39,143 +39,163 @@ leftTemperature,rightTemperature,k,solution,age,batchId,fitness,iteration,accept
...
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
-1690.65387733,1690.65387733,0.0210071821403,1690.6538681,0,9,-0.0210071821403,8,accepted,81.4225770116,,0.292893218813,False
+1690.65387733,1539.41137861,0.0219179489586,1606.63026451,0,3,-0.0219179489586,2,accepted,218.841385912,,0.702642533249,False
```
wangcj05, line 42: The format of this csv file looks weird to me. Could you check it?

tests/framework/Optimizers/GeneticAlgorithms/gold/discrete/constrained/failureOfRuns/opt_export_0.csv
```diff
@@ -44,67 +44,40 @@ x1,x2,x3,ans,age,batchId,fitness,iteration,accepted,AHDp,conv_AHDp
...
+2.0,1.0,2.0,10.0,0.0,9.0,-10.0,8.0,accepted,0.0,1.0
+2.0,1.0,2.0,10.0,0.0,9.0,-10.0,8.0,accepted,0.0,1.0
+1.0,2.0,2.0,11.0,0.0,9.0,-11.0,8.0,accepted,0.0,1.0
+2.0,1.0,2.0,10.0,0.0,9.0,-10.0,8.0,accepted,0.0,1.0
+2.0,1.0,2.0,10.0,0.0,9.0,-10.0,8.0,accepted,0.0,1.0
+1.0,2.0,2.0,11.0,0.0,9.0,-11.0,8.0,accepted,0.0,1.0
+1.0,2.0,2.0,11.0,0.0,9.0,-11.0,8.0,accepted,0.0,1.0
+1.0,2.0,2.0,11.0,0.0,9.0,-11.0,8.0,accepted,0.0,1.0
+2.0,1.0,2.0,10.0,0.0,9.0,-10.0,8.0,accepted,0.0,1.0
+2.0,2.0,1.0,9.0,0.0,9.0,-9.0,8.0,final,0.0,1.0
```
wangcj05, line 83: why in some test cons_AHDp using 1 and 0 to indicate True and False, and in other test True and False are use?

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -314,12 +317,15 @@ def __init__(self):
...
     self.needDenormalized()                                      # the default in all optimizers is to normalize the data which is not the case here
     self.batchId = 0
-    self.population = None                                       # panda Dataset container containing the population at the beginning of each generation iteration
-    self.popAge = None                                           # population age
-    self.fitness = None                                          # population fitness
-    self.rank = None                                             # population rank (for Multi-objective optimization only)
-    self.constraintsV = None                                     # calculated contraints value
-    self.crowdingDistance = None                                 # population crowding distance (for Multi-objective optimization only)
+    self.prevPop_inputs = None
+    self.currentPop_ages = None
```
wangcj05, line 321: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -314,12 +317,15 @@ def __init__(self):
...
+    self.matingPopInputs = None                                 # panda Dataset container containing the population at the beginning of each generation iteration
+    self.matingPopObjVals = None                                # objective values of solutions
+    self.matingPopAges = None                                   # population age
+    self.matingPopFitness = None                                # population fitness
+    self.matingPopRanks = None                                  # population rank (for Multi-objective optimization only)
+    self.matingPop_g = None                                      # calculated contraints value
```
wangcj05, line 327: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -700,6 +712,16 @@ def handleInput(self, paramInput):
...
       self._objCoeff = fitnessNode.findFirst('a').value if fitnessNode.findFirst('a') else None
+    self._normalizeFitness = fitnessNode.findFirst('normalize').value if fitnessNode.findFirst('normalize') else None
+    if not self._normalizeFitness:
+      self._normalizeFitness = False
+    elif self._normalizeFitness.lower() == 'none':
+      self._normalizeFitness = None #allow the user to specify NoneType in the string input
+    elif self._normalizeFitness.lower() == 'true':
+      self._normalizeFitness = 'zscore' #default to zscore normalization
+    elif self._normalizeFitness.lower() not in ['maxmin','zscore']:
+      self.raiseAnError(IOError, "Requested fitness normalization type not supported. Available options include 'maxmin' and 'zscore'.") #!TODO: maxmin has not yet been implemented.
```
wangcj05, line 723: I think you should use makeEnumType in the InputData so that you can avoid the multiple if elif checks here

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
       self._closeTrajectory(t, 'cancel', 'Currently GA is single trajectory', 0)
     self.incrementIteration(traj)
 
-    population = datasetToDataArray(rlz, list(self.toBeSampled))
-
-    objectiveVal = []
+    currentPopInputs = datasetToDataArray(rlz, list(self.toBeSampled))
+    currentPop_objvals = []
```
wangcj05, line 826: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
-    # Single-objective post-processing (if needed)
-    if not self._isMultiObjective:
-        self._collectOptPoint(rlz, populationFitness, objectiveVal[0], g)
-        self._resolveNewGeneration(traj, rlz, info, objectiveVal[0], populationFitness, g)
-
-    # 3. Survivor selection
+      currentPop_objvals.append(list(np.atleast_1d(rlz[self._objectiveVar[i]].data)))
+
+  ## 1. Check constraint violations and calculate the constraint function g (<0 if the constraint is violated)
+    currentPop_g = constraintHandling(self, info, rlz, currentPopInputs, currentPop_objvals, multiObjective=self._isMultiObjective)
```
wangcj05, line 831: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
+
+  ## 2. Normalize values for fitness function evaluation, if requested.
+    norm_rlz = deepcopy(rlz)
```
wangcj05, line 834: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
+        varsToNormalize += x.parameterNames()
+      varsToNormalize = set(varsToNormalize + self._objectiveVar)
+      # calculate normalization parameters and store for later
+      self.normScores = {}
+      for var in varsToNormalize:
+        if self._normalizeFitness == "zscore":
+          self.normScores[var] = (np.mean(rlz[var].to_dataframe().values), np.std(rlz[var].to_dataframe().values))
+          # normalize values for fitness calc
+          for i in range(len(rlz[var])):
+            norm_rlz[var][i] = (rlz[var][i] - self.normScores[var][0]) / self.normScores[var][1] #perform zscore normalization
```
wangcj05, line 849: camelBack

wangcj05: in addition, it is better to have a function for normalization

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
+            if np.isnan(norm_rlz[var][i]):
+              norm_rlz[var][i] = 0.0
+      # normalize evaluated constraint differences
+      for i in range(len(currentPop_g)):
+        for j in range(len(constrVarsList)):
+          #penalty represents a Delta applied to the obj value; i.e. (C_k - mu)/std - (C_i - mu)/std = (C_k - C_i)/std
+          #  If the constraint uses multiple variables/units, there's no easy way to handle it so we'll just assume the
+          #  first variable is the primary one.
+          currentPop_g[i][j] = currentPop_g[i][j] / self.normScores[constrVarsList[j].parameterNames()[0]][1]
```
wangcj05, line 858: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
+                                                         currentPop_objvals,
+                                                         currentPop_g)
+      else:
+        survivorSelectionProcess.multiObjSurvivorSelect(self, info, rlz, traj,
+                                                        currentPopInputs,
+                                                        currentPopFitness,
+                                                        currentPop_objvals,
+                                                        currentPop_g)
+      # update the age estimate for the current population for printing later
+      self.currentPop_ages = np.zeros(len(currentPopInputs), dtype=int)
```
wangcj05, line 893: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -796,90 +822,154 @@ def _useRealization(self, info, rlz):
...
+      # initialize multi-objective containers
+      currentPopRanks = []; currentPopCD = [] #!TODO: it would be better to simply call _parentSelectionInstance two different ways than require these values be initialized.
+  ## Single-objective post-processing
+      if not self._isMultiObjective:
+          self._collectOptPoint(rlz, currentPopFitness, currentPop_objvals[0], currentPop_g)
+          self._resolveNewGeneration(traj, rlz, info, self.prevPop_inputs, currentPop_objvals[0], currentPopFitness, currentPop_g)
+  ## Multi-objective post-processing
+      else:
+        # list for Rank and CD calculation
+        currentPop_fitsbysoln = datasetToDataArray(currentPopFitness, self._objectiveVar).data.tolist()
```
wangcj05, line 911: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -902,13 +992,23 @@ def _useRealization(self, info, rlz):
...
           newRlz[var] = float(daChildren.loc[i, var].values)
-        self._submitRun(newRlz, traj, self.getIteration(traj))
+
+        childrenToSubmit.append(newRlz)
+
+      for child in childrenToSubmit:
+        self._submitRun(child, traj, self.getIteration(traj))
+
+  ## 11. Save grandparents
+    self.prevPop_inputs = deepcopy(currentPopInputs)
```
wangcj05, line 1011: camelBack

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -983,21 +1183,25 @@ def _resolveNewGeneration(self, traj, rlz, info, objectiveVal=None, fitness=None
...
           for ind, consName in enumerate([y.name for y in (self._constraintFunctions + self._impConstraintFunctions)]):
-            rlzDict['ConstraintEvaluation_'+consName] = self.constraintsV.data[i,ind]
+            rlzDict['ConstraintEvaluation_'+consName] = g.data[i,ind]
         else:
           varList = self._solutionExport.getVars('input') + self._solutionExport.getVars('output') + list(self.toBeSampled.keys())
           rlzDict = dict((var,np.atleast_1d(rlz[var].data)[i]) for var in set(varList) if var in rlz.data_vars)
           for j in range(len(self._objectiveVar)):
-            rlzDict[self._objectiveVar[j]] = np.atleast_1d(rlz[self._objectiveVar[j]].data)[i]
+            if self._objectiveVar[j] not in rlzDict:
+              rlzDict[self._objectiveVar[j]] = np.atleast_1d(rlz[self._objectiveVar[j]].data)[i]
```
wangcj05, line 1204: Does these lines mean the objective value will not be updated if the it is already in rlzDict? Could you verify to see if this will cause any problem or not?

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -1053,7 +1258,7 @@ def _collectOptPoint(self, rlz, fitness, objectiveVal, g):
 
     return point
 
-  def _collectOptPointMulti(self, population, rank, CD, objVal, fitness, constraintsV):
+  def _collectOptPointMulti(self, rlz, population, rank, CD, objVal, fitness, constraintsV):
```
wangcj05, line 1261: update docstring for rlz?

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -696,6 +1285,12 @@ def _updateSolutionExport(self, traj, rlz, acceptable, rejectReason):
         toExport[var] = rlz[var]
     # formatting
     toExport = dict((var, np.atleast_1d(val)) for var, val in toExport.items())
+    # Force solutionExport to expect all the vars we want to give it.
+    for key in toExport.keys():
+      if key not in self._solutionExport.vars:
+        self._solutionExport.addedVars.append(key)
+    self._solutionExport.addedVars = list(set(self._solutionExport.addedVars))
```
wangcj05, line 1292: If we need to force the vars into solutionExport, it seems we did not handle it properly.

wangcj05 (changes requested):
@W0lfShAd0w I have some comments for you to consider. In addition, I need justification for the regold files, some have different convergence, some have different lengths, and some have different values.

ravenframework/Optimizers/fitness/fitness.py
W0lfShAd0w: Adding a negative penalty is more logically consistent with the Genetic Algorithm method than subtracting a positive penalty. It is unnecessary to ask every new developer to perform the mental gymnastics required to understand a nontraditional approach to applying fitness penalties.

That being said, these concepts are ultimately mathematically indistinguishable.

ravenframework/Optimizers/fitness/fitness.py
W0lfShAd0w: This might be a misrepresentation of the git diff? I don't think the default was already changed to 1.0.
The old default was 10.0, I think that is negligently misleading and the default SHOULD instead be changed to 1.0. However, I reversed the change until I could run it by you first.

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: This has been replaced with a more generic solution.

ravenframework/Optimizers/BayesianOptimizer.py
W0lfShAd0w: The raiseADebug message was NOT removed, just migrated to the shared method "_queueSubmission".

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/mishraBirdConstrainedInvLin/opt_export_0.csv
W0lfShAd0w: So, I thought I fixed that. This "0.0 for all entries in the age column" appears to be an old bug that was golded over a few years ago and never caught. The issue was related to how children and parents are compared to identify if the "age" value should be incremented. Ultimately, I still think our current approach is fundamentally flawed. I asked Claude to make a comparison across the fix and weigh in:

Can we be certain the new values are more accurate?

No — not in these specific rows, and I should be precise about why. Once several tracked individuals are literally indistinguishable (identical genes, no separate lineage ID kept anywhere in the code), "the age of this individual" is not a well-defined quantity — there is no information in the data model that says which of the 26 duplicate lineages the just-evaluated point "really" came from. Both approaches are picking one candidate out of a genuinely ambiguous set:

- Old (.equals()): picks whichever mating-pool entry happens to share the same position/coordinate label as the current entry — an incidental, opaque byproduct of how the two arrays happened to be ordered, not a deliberate rule.
- New (np.array_equal): picks the first value-match in iteration order — deterministic and simple, but still an arbitrary choice among equally-valid candidates (and since fitnessBased sorts duplicates newest-first on ties, this systematically tends toward the youngest candidate, e.g. picking 0 in both examples above).

Neither is "the correct" answer; they're different tie-breaks over an inherently ambiguous case. What I can say with certainty:

- Everywhere the match is unambiguous (a unique genotype, or duplicates that all happen to share one age) — which is the overwhelming majority of rows in all 16 files — the new value is provably correct and the two methods agree (hence 0 diff there).
- The fix is still worth keeping: it correctly fixes the cases that were unambiguous but previously failed outright (like mishraBirdConstrainedInvLin, where matches almost never succeeded even for non-duplicated points). It just doesn't — and can't — resolve the deeper ambiguity that surfaces once the population has genuinely collapsed to duplicate values.

If you want the ambiguous cases to have a real, principled answer rather than an arbitrary pick, that requires tracking a persistent per-chromosome lineage ID through crossover/mutation/survivor-selection, not just genotype-value matching — a materially bigger change than this reporting fix, and a separate design decision.

Neither is "the correct" answer; they're different tie-breaks over an inherently ambiguous case.

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/rosenbrockDiskConstrained/opt_export_0.csv
W0lfShAd0w: see above conversation.

ravenframework/Optimizers/GeneticAlgorithm.py
W0lfShAd0w: It doesn't cause any problems. It prevents a problem by stopping an overwrite of information that is stored in the dict from the line just above "rlzDict = dict((var,np.atleast_1d(rlz[var].data)[i]) for var in set(varList) if var in rlz.data_vars)".

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: I wholeheartedly agree. I expressed this criticism late last Fall: the way we are arbitrarily restricting the definition of which vars should go into the _solutionExport is silly. This was my minimalist workaround for ensuring all relevant vars are passed along through the _solutionExport.

W0lfShAd0w:
@wangcj05 The reviewer comments have been addressed. Please let me know if additional changes are needed.

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -421,10 +860,157 @@ def flush(self):
...
+    """
+    key = []
+    for var in sorted(self.toBeSampled.keys()):
+      val = point[var]
+      if hasattr(val, 'values'):
+        val = val.values
+      if hasattr(val, 'data'):
+        val = val.data
+      arr = np.asarray(val).reshape(-1)
+      key.append(tuple(int(x) for x in arr))
```
wangcj05, line 884: [blocking] `_makeSubmissionKey` truncates values with `int(x)`, so dedup silently corrupts continuous optimization.

```python
key.append(tuple(int(x) for x in arr))
```

`deduplication` is documented as applying to *all* RavenSampled optimizers, but `int(x)` collapses every float that shares an integer part. For continuous variables (GradientDescent/SimulatedAnnealing/BayesianOptimizer, and continuous GA like `mishraBird`) — often normalized to [0,1] — essentially every point maps to the same key, so after the first evaluation every subsequent point is skipped as a "duplicate" and replayed from cache. That's silently wrong science with no warning.

What do you think about either (a) keying on the exact (or explicitly rounded) values, or (b) restricting the feature to discrete variables and `raiseAnError` otherwise? At minimum the input doc shouldn't claim universal applicability.

ravenframework/Models/Code.py
```diff
@@ -835,9 +836,10 @@ def evaluateSample(self, myInput, samplerType, kwargs):
 
       ## Check if the user specified any file extensions for clean up
       for fileExt in fileExtensionsToDelete:
-        fileList = [ os.path.join(metaData['subDirectory'],f) for f in os.listdir(metaData['subDirectory']) if f.endswith(fileExt) ]
-        for f in fileList:
-          os.remove(f)
+        pattern = '*.'+fileExt.removeprefix('.').removeprefix('*.')
+        for path in Path(metaData['subDirectory']).rglob(pattern):
+            if path.is_file():
+                path.unlink()
```
wangcj05, line 842: [blocking] `Code.py` cleanup became recursive — wider deletion blast radius.

Two behavior changes from the old `os.listdir(...) + endswith`: it now recurses into **all** subdirectories via `rglob` (previously top-level only), and `endswith(ext)`→`*.ext` glob changes matching semantics. The recursion is the concern — if a run's subdirectory contains nested outputs that happen to match a cleanup extension, they'll now be deleted where they previously survived.

Is the recursive scope intentional? If so, could you call it out in the PR description; if not, use `glob` (non-recursive) to preserve the original scope.

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -385,8 +779,53 @@ def finalizeSampler(self, failedRuns):
...
+        objectiveVars.append(obj)
+      else:
+        matches = [var for var in optKeys if obj.lower() in var.lower()]
+        objectiveVars.extend(matches)
+    # Deduplicate while preserving order
+    objectiveVars = list(dict.fromkeys(objectiveVars))
+
+    self._solutionExport.asDataset() #empty _collector into _data
+    ## Search the rest of the history
+    if hasattr(self,"_sampledHistoryInfo"):
```
wangcj05, line 799: [important] Dead code in `finalizeSampler` — attribute name mismatch.

The "search the rest of the history for better solutions" block is guarded by `if hasattr(self, "_sampledHistoryInfo"):`, but `_sampledHistoryInfo` is never assigned anywhere — survivorSelection populates `_sampledPopulationInfo` (different name). So this block never runs. If it's meant to be live, it's wired to the wrong attribute; if not, could it come out? It also reaches into `self._solutionExport._data` directly and has an acknowledged "would fail silently" path. Relatedly, `self._solutionExport.asDataset()` is now called unconditionally here for all optimizers — is that side effect intended?

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -421,10 +860,157 @@ def flush(self):
...
+      Caches evaluated points so later duplicate submissions can be skipped.
+      @ In, rlz, dict or xr.Dataset, normalized realization(s)
+      @ Out, None
+    """
+    if not self._deduplication:
+      return
+    if isinstance(rlz, dict):
+      key = self._makeSubmissionKey(rlz)
+      self._evaluatedSubmissionKeys.add(key)
+      self._evaluatedSubmissionData[key] = copy.deepcopy(rlz)
```
wangcj05, line 898: [important] Unbounded dedup cache + heavyweight checkpoints.

`_evaluatedSubmissionData` keeps a `deepcopy` of every evaluated realization for the entire run, and `_getCheckpointState` serializes that whole cache to JSON on every write. For long runs (the checkpoint feature's whole motivation), memory and per-checkpoint IO grow with run length. Could we bound the cache or exclude it from the checkpoint?

ravenframework/Optimizers/fitness/fitness.py
```diff
@@ -47,24 +47,31 @@ def invLinear(rlz, **kwargs):
...
         b, list of floats, linear coefficient(s) for the penalty measure (default = 10.0 for each objective)
         constraintFunction, xr.DataArray, measuring the severity of the constraint violation.
         type, list of strings, indicating 'min' or 'max' for each objective
   @ Out, fitnessSet, xr.Dataset, the fitness function for the given population.
   """
   objVar = kwargs['objVar']
-  a = [_defaultObjectiveScaling] * len(objVar) if kwargs.get('a') is None else kwargs['a']  # Scaling factors for objectives
-  b = [_defaultPenaltyScaling] * len(objVar) if kwargs.get('b') is None else kwargs['b']  # Penalty scaling factors
-  if len(a) != len(objVar):
+  constraintNum = kwargs['constraintNum']
```
wangcj05, line 62: [important] Fitness functions now hard-require `constraintNum`.

`invLinear` here (and `feasibleFirst` at line 120) do `kwargs['constraintNum']` with no default, so any caller/plugin not passing it now gets a `KeyError` instead of the old default behavior. The new `b`-reshape logic is also intricate and only exercised via gold diffs. A direct unit test on `invLinear`/`feasibleFirst` (there isn't one today) would protect this.

wangcj05 (commented):
## Code review — PRLO modifications to Optimizer-related classes

**Review depth: Deep.** New subsystem (checkpoint/restart) plus a cross-cutting dedup feature touching every `RavenSampled` optimizer, with a large GA rewrite. The ~58-file count is misleading — most are regenerated gold CSVs; the real change is ~1,100 lines across 17 source files.

Up front: this PR bundles several independent features — checkpoint/restart, submission deduplication, fitness normalization, a GA population/aging rewrite, and unrelated fixes to `Code.py`/`Simulation.py`/`randomUtils.py`. That mix makes it hard to review and hard to bisect later. Would you consider splitting at least checkpoint/restart and dedup into separate PRs?

### [blocking] CI is red on every platform — the PR fails its own regression suite
`test_GA_PARCS_LP` and `Optimizers/GeneticAlgorithms/mishraBirdConstrained` fail on Linux/Windows/macOS with `age` mismatches ("Different values in age for 0.0 and 2.0", etc.). The gold CSVs were regenerated in this PR, yet the `age` column still doesn't match the code's output. Combined with the head commits ("Failed to make age increment properly", "Previous age parameter fix wasn't making it to the CSV", "age values became reciprical and always 0.0"), the `age` computation looks unstable/non-deterministic. This needs to be green before merge.

### [important] No tests for any of the new features
The only test changes are two kwarg renames + regenerated golds. Checkpoint write/restore, deduplication, and fitness `normalize` have zero direct coverage. For checkpoint/restart specifically, a small round-trip unit test (write state → restore → assert equality; plus a validation-failure case for mismatched population size / optimizer type) would catch a lot. Given the JSON/ast/h5py encode-decode machinery, this is where I'd most want tests before merge.

---
**Bottom line:** Ambitious work — the checkpoint/restart design (versioned, validated, JSON-safe encoding) is thoughtful. But it's not merge-ready: CI is failing on the age output, dedup has a silent correctness trap for continuous variables, and the new subsystems ship without tests. I'd get CI green, fix or gate the dedup key, wire up or remove the dead history block, and add round-trip tests before another pass.

ravenframework/DataObjects/DataSet.py
```diff
@@ -65,7 +65,8 @@ def __init__(self):
     self._neededForReload = [self.sampleTag] # metavariables required to reload this data object.
     self._samplerTag      = None
     self.inputKDTree      = None
-    self._autogenerate    = set()             # index vars in here are automatically generated
+    self._autogenerate    = set()            # index vars in here are automatically generated
+    self.addedVars        = []               # placeholder list for additional vars to be explicitly requested. This is not intended for general use.
```
wangcj05, line 69: [important] `DataSet.addedVars` is a fragile cross-object hack.

The optimizer appends to this public `addedVars` list on the DataObject, then does `= list(set(...))` — which reorders columns nondeterministically — and the list persists across realizations. The comment itself says "not intended for general use." Could the needed vars be declared through the existing variable/meta-key machinery instead?

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -321,10 +476,249 @@ def localFinalizeActualSampling(self, jobObject, model, myInput):
...
+          self._solutionExport.addRealization(row)
+        self.raiseAMessage(f'Restored {len(rows)} SolutionExport row(s) from restart file.')
+      except Exception as err:
+        self.raiseAWarning(f'Could not restore SolutionExport data from restart file: {err}')
+    self._checkpointRestored = True
+    gen = checkpoint.get('generation', 'unknown')
+    self.raiseAMessage(f'Successfully resumed optimizer "{self.name}" from restart file '
+                      f'"{self._restartFromFile}" (last completed generation: {gen}).')
+
+  def finalizeSampler(self, failedRuns): #!TODO: is this unused??
```
wangcj05, line 721: [important] Author-uncertainty comment left in production code.

`#!TODO: is this unused??` — it is used; the "Final Optimal Point" logic lives inside `finalizeSampler`. Similar `#!TODO` markers and `- rollnk` / `#NOTE(rollnk)` signatures appear in `DataSet.py`, `GeneticAlgorithm.py`, `survivorSelection.py`, and `Simulation.py`. Could you resolve these or convert them to tracked issues, and drop the personal signatures?

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -458,7 +464,7 @@ def getInputSpecification(cls):
                     \item \textit{uniformCrossover} - It randomly selects genes from two parent chromosomes with equal probability, creating offspring by exchanging genes at corresponding positions.
                   \end{itemize}""")
     crossover.addParam("type",
-                       InputTypes.makeEnumType('crossover','crossoverType',['onePointCrossover','twoPointsCrossover','uniformCrossover']),
+                       InputTypes.makeEnumType('crossover','crossoverType',crossoversList),
```
wangcj05, line 467: [nit] `makeEnumType(..., crossoversList)` passes the `__crossovers` *dict* where the signature documents a "list of strings". It works (consumers only iterate / `in` the keys), but the name and comment ("updated list") are misleading — consider `list(crossoversList)`.

W0lfShAd0w:
Issue raised: #2586
Failed test on the CentOS 7 machine is preexisting in devel with intermittent behavior. I've raised an issue to catalog this so it can be fixed in a separate PR (issue is unrelated to this PR).

ravenframework/Models/Code.py
W0lfShAd0w: This change was necessary to support a code interface workflow that generates data in multiple subdirectories without drastically inflating the size-on-disk requirements for RAVEN. I've added a toggle (i.e. recursive=True) and changed the default behavior back to being non-recursive.

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: Fixed by option (a).

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: This was renamed in a effort to be more general, but I think the more appropriate fix would be to guard this search to be GA-only in the first place.

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: Fixed dedup key truncation, revived dead GA history search, cap dedup cache size.

ravenframework/Optimizers/fitness/fitness.py
W0lfShAd0w: This feels like a question of architectural preference. Let me know if you want it changed/adjusted.

ravenframework/DataObjects/DataSet.py
W0lfShAd0w: Fixed: removed addedVars for a hardened equivalent.

ravenframework/Optimizers/GeneticAlgorithm.py
W0lfShAd0w: Fixed this for crossovers and for mutators.

ravenframework/Optimizers/RavenSampled.py
W0lfShAd0w: resolved open comments and signatures.

ravenframework/Optimizers/fitness/fitness.py
wangcj05: No need to change for this PR.

ravenframework/Optimizers/fitness/fitness.py
wangcj05: I think we should use 1.0.

tests/framework/Optimizers/GeneticAlgorithms/gold/continuous/constrained/mishraBirdConstrainedInvLin/opt_export_0.csv
wangcj05: I guess we should remove the age in the output for now.

ravenframework/Optimizers/RavenSampled.py
wangcj05: ok, please raise an issue for this.

W0lfShAd0w:
Added unit tests to cover new functionality added in this PR, per the reviewer's comments.

ravenframework/Optimizers/GeneticAlgorithm.py
wangcj05: check this comment

ravenframework/Optimizers/GeneticAlgorithm.py
wangcj05: check this comment

ravenframework/Optimizers/GeneticAlgorithm.py
wangcj05: same here

ravenframework/Optimizers/GeneticAlgorithm.py
wangcj05: check here

ravenframework/Optimizers/GeneticAlgorithm.py
```diff
@@ -843,12 +875,11 @@ def _useRealization(self, info, rlz):
...
         if self._normalizeFitness == "zscore":
-          self.normScores[var] = (np.mean(rlz[var].to_dataframe().values), np.std(rlz[var].to_dataframe().values))
+          mean, std, normalizedValues = self._computeZscoreNormalization(rlz[var].to_dataframe().values)
+          self.normScores[var] = (mean, std)
           # normalize values for fitness calc
           for i in range(len(rlz[var])):
-            norm_rlz[var][i] = (rlz[var][i] - self.normScores[var][0]) / self.normScores[var][1] #perform zscore normalization
-            if np.isnan(norm_rlz[var][i]):
-              norm_rlz[var][i] = 0.0
+            norm_rlz[var][i] = normalizedValues[i]
```
wangcj05: camelBack

ravenframework/Optimizers/RavenSampled.py
```diff
@@ -563,7 +581,7 @@ def _validateCheckpoint(self, checkpoint):
       @ In, checkpoint, dict, loaded checkpoint dict
       @ Out, None
     """
-    currentVersion = '1.0' # Currently supported checkpoint version #NOTE: update this if the checkpoint format is changed.
+    currentVersion = _CHECKPOINT_VERSION # Currently supported checkpoint version.
```
wangcj05: For future, I guess it is better to have a list of supported version instead of scalar value here

wangcj05 (changes requested):
some minor comments, please try to check the variables names, several of them are still using underscore. @W0lfShAd0w

wangcj05 (approved):
changes are good.

W0lfShAd0w:
Results from PRLO test suite:
Rook test results:
image

Pytest results:
image

PR #2589, opened 2026-09-17
Optimization plotting catalogue

Jimmy-INL:
--------
Pull Request Description
--------
##### What issue does this change request address? (Use "#" before the issue to link it, i.e., #42.)

Closes #2544

##### What are the significant changes in functionality due to this change request?

This change request extends the OutStreams optimization-plotting catalogue with a
generation-inspection capability and consolidates the parallel-coordinate plots.

- **New shared helper `plotGenerationUtils`.** Centralizes the generation
  down-sampling and PNG frame-selection logic that had been duplicated across the
  animated plotters (`_sample_generations` / `_select_frame_indices`).

- **Optional `<generations>` selector** wired into `DiversityRadarPlot`,
  `ConstraintActivityTimelinePlot`, `SamplingCoverageMapPlot`,
  `HypervolumeMoviePlot`, `TradeoffSlicePlot`, and `ObjectiveContourAnimationPlot`.
  When supplied, the plotter renders exactly the requested generations (explicit
  batch ids such as `5`, the keywords `first`/`last`, and inclusive ranges
  `start:stop` / `start:stop:step`) and saves them all as PNGs. Requested
  generations absent from the data raise an error. When the node is omitted, the
  historical even-sampling behavior (`max_frames`, `frames_max`) is unchanged, so
  the quick-overview default is preserved.

- **Merged parallel-coordinate subtype.** `OptParallelCoordinatePlot` now renders
  the classic one-frame-per-generation beams by default and switches to the
  enhanced constraint-aware / trail / generation-selection rendering when any
  enhancement node is provided. The standalone `OptParallelCoordinateConstraintPlot`
  subtype is removed and folded into this single subtype.

- **Tests and documentation.** Adds two regression tests
  (`NSGA-II_ZDT1_plots`, `NSGA-II_MinwoRepMultiObjective_plots`) that exercise the
  full plotting catalogue, the merged parallel subtype, and the explicit
  `<generations>` selector; updates `OptimizationPlots.md` to describe the merged
  subtype and mark the generation-inspection feature as implemented.

Additive only: no existing plot subtype behavior changes by default.

----------------
For Change Control Board: Change Request Review
----------------
The following review must be completed by an authorized member of the Change Control Board.
- [x] 1. Review all computer code.
- [x] 2. If any changes occur to the input syntax, there must be an accompanying change to the user manual and xsd schema. If the input syntax change deprecates existing input files, a conversion script needs to be added (see Conversion Scripts).
[...]

PR #2590, opened 2026-09-18
Vectorize DataSet scaling-factor computation

Jimmy-INL:
## Summary
Rework `DataSet._setScalingFactors` so it:
- restricts to **numeric** variables (skips strings/objects) via a cheap dtype metadata check, and
- masks **non-finite** entries (NaN and ±inf) before computing per-variable mean/scale,

while performing the reduction with **xarray's vectorized** `mean`/`std` instead of a per-variable Python loop. The only remaining loop just reads two precomputed scalars per variable, so the cost stays flat as the number of variables grows.

## Why
Addresses review feedback that a per-variable Python loop over `ds.data_vars` scales poorly for data objects with many variables. This keeps the correctness guards (numeric-only, finite-only) while removing the Python-level per-variable iteration.

## Behavior
- Equivalent scale factors to the guarded loop; an all-non-finite variable yields NaN mean/scale.
- Restores the vectorized approach previously in `devel`, now with the added numeric/finite guards.

## Scope
Single-file change: `ravenframework/DataObjects/DataSet.py`.


Jimmy-INL:
Superseded — the vectorized `_setScalingFactors` fix now lives on #2589 (the branch where @wangcj05's review comment was made), so there's no need for a separate PR. Closing this.

